Issue #6, opened 2016-09-29
cannot compile hlc


andyli:
```
Error	LNK2019	unresolved external symbol _hl_entry_point referenced in function _wmain	hlc	D:\Documents\workspace\hl\hlc_main.obj	1
Error	LNK2019	unresolved external symbol _hlc_static_call referenced in function _wmain	hlc	D:\Documents\workspace\hl\hlc_main.obj	1
Error	LNK2019	unresolved external symbol _hlc_get_wrapper referenced in function _wmain	hlc	D:\Documents\workspace\hl\hlc_main.obj	1
Error	LNK1120	3 unresolved externals	hlc	D:\Documents\workspace\hl\Debug\hlc.exe	1
```

In hlc.vcxproj, there is a reference to `src\_main.c`, but it doesn't exist in the repo. Maybe that's the reason?

Simn:
I think you're supposed to compile _your_ code as `_main.c`.

ncannasse:
I have not yet completed x64 support. Can you try with -m32 for now ?

Le 29 sept. 2016 8:40 AM, "Simon Krajewski" <notifications@github.com> a
écrit :

andyli:
Simon:
Ah-ha!
Then I guess it makes sense to exclude the hlc project in the VS solution by default.

Nicolas:
Will try that. You're commenting on the wrong issue ;)

ncannasse:
Sorry for the wrong issue, I should not try to use mobile phones :)

ncannasse:
I wonder how we should distribute this. We could have a static  hl_runtime static lib to link together with _main.c , this will only require the hl.h and hlc.h headers. For Windows that's for sure, I wonder for Linux if we should let the users do the full compilation themselves or distribute precompiled libs

andyli:
Let's let Linux users do the full compilation. It is just too hard to build truly portable Linux binaries. When we have a proper linux packaging for hl, the Linux distros will automatically provide the precompiled binaries.

andyli:
Do you have an existing _main.c that can be used for testing hlc?

ncannasse:
```
class Main {
    static function main() trace("Hello world!");
}
````

`haxe -hm _main.c -main Main`

ncannasse:
HLC should be compilable in both 32 and 64 bits now

PR #7, opened 2016-09-29
Remove hlc project from the build config. (see #6)

andyli:
P.S. hlc is still there in the solution. But it just wouldn't be built when building the solution.


ncannasse:
I need it for my workflow, and it's not much a problem since for windows users we will distribute precompiled libs + headers so they don't have to compile HL by themselves.

Issue #9, opened 2016-10-21
Program never completed


eliasku:
I tried to run tests for `hxease`, and running with `hl test/test.hl` starts and never completes with 100% cpu usage.

I use OSX 10.11.5.
`haxe` nightly  from 20 oct 2016
`hl` builded for osx with `make` from git at 20 oct 2016 (as how it's done on hl travis-ci script)
hxease: https://github.com/eliasku/hxease/tree/develop

Builded tests which is never completes: [test.hl.zip](https://github.com/HaxeFoundation/hl/files/543636/test.hl.zip)

ncannasse:
What is the source code for the example?

ncannasse:
(or can it be reproduced with a smaller pure Haxe code sample without any lib?)

eliasku:
@ncannasse I will try to pack repro in 30 minutes for you

eliasku:
[hl-repro.zip](https://github.com/HaxeFoundation/hl/files/545273/hl-repro.zip)
There are `hl.hxml` and `neko.hxml`

haxiomic:
The repro is also hanging on my mac running 10.11.6

I'm having the same issue running the mandlebrot test, it gets to pixel (214,1) and hangs in both hl JIT and compiled HLC

ncannasse:
I can't reproduce on Windows, I'll have to check on OSX at work this week.

haxiomic:
I've made a PR for the fix and the hl-repro project no longer hangs

It turned out to be the TRAILING_ZEROES macro. On windows it's defined as an inline function whereas on other platforms it's the macro:
```c
#define TRAILING_ZEROES(x)	(x?__builtin_ctz(x):32)
```

The problem is if x is a ternary expression like `a ? b : c` then if a is true TRAILING_ZEROES will return a rather than __builtin_ctz(a).

I've changed the macros to inline functions which also prevents x being evaluated twice.

The hang itself occurred in gc_alloc_var - [alloc.c:383](https://github.com/HaxeFoundation/hl/blob/master/src/alloc.c#L383) bits gets set incorrectly because of the ternary

PR #11, opened 2016-10-22
Fix hang on mac and linux

haxiomic:
The TRAILING_ZEROES macro did not ensure correct evaluation order (it behaved incorrectly when x was a ternary expression for example). To fix this I've replaced these macros with inline functions. Adding brackets around the first x in the TRAILING_ZEROES macro would have worked as well, but we're probably better off using an inline function in this case.

Resolves issue #9

### DIFF
--- a/src/alloc.c
+++ b/src/alloc.c
@@ -38,8 +38,12 @@ static unsigned int __inline TRAILING_ZEROES( unsigned int x ) {
 #else
 #	include <sys/types.h>
 #	include <sys/mman.h>
-#	define TRAILING_ONES(x)		(~(x)?__builtin_ctz(~(x)):32)
-#	define TRAILING_ZEROES(x)	(x?__builtin_ctz(x):32)
+static inline unsigned int TRAILING_ONES( unsigned int x ) {
+	return (~x) ? __builtin_ctz(~x) : 32;
+}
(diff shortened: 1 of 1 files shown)


ncannasse:
I trust you on this, but I'm curious why/how this makes a difference?

Simn:
The textual replacement gives you some `a ? b : c ? d : 32` construct which probably doesn't evaluate like you want it to.

Also, macros should not use their argument values multiple times because this blows up when calling it with `c++` or some other side effect.

ncannasse:
@Simn yes I was careful not to do any side effect in arguments. 
I spoted the problm, it could have been fixed by changing
`TRAILING_ZEROES(x)	(x?__builtin_ctz(x):32)` by
`TRAILING_ZEROES(x)	((x)?__builtin_ctz(x):32)`
But it's ok to use inline function as well. I hope GCC does respect it.

Simn:
Given that inline functions are the go-to replacement for "evil" macros, I wouldn't worry about this.

haxiomic:
I find this a good tool to answer questions of which approach is better

https://godbolt.org/g/bn7S0P

The coloured regions show the regions of code responsible for regions of assembly.

And it'll let you compile it with a large range of compliers and versions - super useful tool if you've not seen it before

PR #12, opened 2016-10-25
MAP_ANONYMOUS not valid on Yosemite

nanjizal:
MAP_ANONYMOUS is not supported on Yosemite, but not a problem on xcode beta3 Sierra. See for details:
https://github.com/leftmike/foment/issues/4

### DIFF
--- a/src/alloc.c
+++ b/src/alloc.c
@@ -866,7 +866,10 @@ void hl_free( hl_alloc *a ) {
 
 HL_PRIM void *hl_alloc_executable_memory( int size ) {
 #ifdef __APPLE__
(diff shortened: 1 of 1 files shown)


nanjizal:
On hello world I get error ( but worked fine on Sierra setup ) will try the other pull.

> No more data
> Error: Command failed with error 1

andyli:
Can we only define it when it is not defined?

nanjizal:
deleted all haxe and got nightly again now works.

nanjizal:
closing changed to only define if required with new pull request.

PR #13, opened 2016-10-25
MAP_ANONYMOUS not valid on Yosemite, define only if needed.

nanjizal:


nanjizal:
Andy Li is this better?

nanjizal:
I think if missing on linux you would need alternate remap so it's better behind an if apple flag.

andyli:
using `#ifndef` would be slightly better ;)

nanjizal:
ah ok, amended, can't check the code on yosemite as I have now upgraded so likely no longer an issue for me but hopefully makes code more robust.

nanjizal:
amended.

PR #21, opened 2016-11-08
Sdl stencil

pperidont:


libs/sdl/gl.c
```diff
@@ -108,6 +108,13 @@ HL_PRIM void HL_NAME(gl_pixel_storei)( int key, int value ) {
 	glPixelStorei(key, value);
 }
 
+HL_PRIM vbyte *HL_NAME(gl_get_string)(int name) {
+	GLOG("%d", name);
+	const char *r = glGetString(name);
+	vbyte *ret = hl_copy_bytes((vbyte*)r, (int)strlen(r) + 1);
+	return ret;
+}
```
ncannasse: Any reason to copy the bytes ? we will use String.fromUTF8 anyway after that

pperidont:
I deleted the useless hl_copy_bytes() and synced with latest hl changes

PR #25, opened 2016-11-10
Makefile: set LBITS for 32/64 bit build

tong:
Don't wanna argue over if 32|64 bit should be the default target, but this sets and uses the LBIT flag to determine the system.
NOT tested on mac/win.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,16 +1,15 @@
+
+LBITS := $(shell getconf LONG_BIT)
+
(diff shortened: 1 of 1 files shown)


ncannasse:
The problem is that atm HL/JIT is only supported on 32 bits so we need a way to force a 32 bits build. Maybe setting ARCH to LBITS only if it's not defined?

PR #26, opened 2016-11-10
Only call UI on windows

ibilon:
Since it's only available there.


ncannasse:
I'm planning to support UI in OSX/Linux, I'm leaving the PR opened but unmerged until then

ibilon:
In the meantime it allows things to run when using it, otherwise it doesn't start because "couldn't find ui.hdll".

PR #29, opened 2016-11-10
Updated sqlite lib

ConstNW:


ncannasse:
hl_error_msg uses a fixed size buffer, please don't use it with a variable length error message that might overflow it

libs/sqlite/sqlite.c
```diff
@@ -0,0 +1,315 @@
...
+	hl_error_msg(USTR("SQLite error: %s"), hl_buffer_content(b, NULL));
+}
+
+static void HL_NAME(finalize_request)(sqlite_result *r, bool exc ) {
+	r->first = 0;
+	r->done = 1;
+	if( r->ncols == 0 )
+		r->count = sqlite3_changes(r->db->db);
+	if( sqlite3_finalize(r->r) != SQLITE_OK && exc )
+    hl_error("SQLite error: Could not finalize request");
```
ncannasse: tabs please :)

libs/sqlite/std/hl/_std/sys/db/Sqlite.hx
```diff
@@ -0,0 +1,272 @@
...
+{
+	public static function connect( path : hl.types.Bytes ) : SqliteConnectionHandle { return null; }
+	public static function close( c : SqliteConnectionHandle ) : Void { }
+	public static function request( c : SqliteConnectionHandle, sql : hl.types.Bytes ) : SqliteResultHandle { return null; }
+	public static function last_id( c : SqliteConnectionHandle ) : Int { return 0; }
+	public static function result_next( c : SqliteResultHandle ) : hl.types.NativeArray<Dynamic> { return null; }
+	public static function result_get( c : SqliteResultHandle, n : Int ) : hl.types.Bytes { return null; }
+	public static function result_get_int( c : SqliteResultHandle, n : Int ) : Int { return 0; }
+	public static function result_get_float( c : SqliteResultHandle, n : Int ) : Float { return .0; }
+	public static function result_get_length( c : SqliteResultHandle ) : Int { return 0; }
```
ncannasse: result_get_int/float does not seem to be correctly typed, they should be Null no ? (or change the bindings)
result_get should return Dynamic too

ConstNW:
> hl_error_msg uses a fixed size buffer, please don't use it with a variable length error message that might overflow it

hl_error_msg uses uvsprintf, if you change it to uvszprintf then it will be safe from overflow.
What do u think?

ncannasse:
We don't have uvszprintf implemented atm, you also also use a traditional sprintf then hl_error on the result

ncannasse:
Just a final change : could you remove sys/db/Sqlite from the branch and instead submit it from inclusion directly in Haxe ? For database we don't require installing extra haxelibs

ConstNW:
PR: https://github.com/HaxeFoundation/haxe/pull/5822

PR #48, opened 2017-01-24
Improve Mac compat

pperidont:
Adds bindings for `glBindFragDataLocation `and `glGenVertexArrays`
Uses OpenGL core profile on OSX (OpenGL 3+)
Includes ssl.hdll compilation in Makefile


andyli:
Thanks for looking into this!
Would you work on the CMake config instead of the Makefile, which I plan to remove as soon as possible ;)

ncannasse:
@andyli please do not remove any file :) I prefer to have CMake + Makefile

andyli:
I thought we agreed to have only CMake + VS project files. No?

ncannasse:
@andyli I think we can keep the Makefile for most common linux distributions + osx and rely on cmake only for more exotic linux or bsd variants, because while cmake build process works it produces very hard to modify output, making it not very usable for debugging

andyli:
I don't think keeping 3 independent build systems is the way to go.

> produces very hard to modify output, making it not very usable for debugging

I don't understand very well. Did you mean modifying the generated Makefiles?
I suppose you can work on the CMakelists.txt directly, analogue to working on .hx files instead of modifying the compiled .js files.

PR #54, opened 2017-03-05
Use environment's default locale on Windows

### LINKED ISSUE #53: hl_sys_print garbled with Chinese Characters in Windows environment
I think that the hl_sys_print do not need convert "ucs2" to "utf8". In Asian countries, the character encoding of the console is not utf8. 

```haxe
import hl.UI;

class App{
	static function main() {
		var s = "你好!";
		trace(s);  // bad
		UI.dialog("", s, cast 0);  // good
	}
}
```

I got a simple example, it can get correct output(only test in msvc2013 and cygwin's gcc).

test.c [save as utf-8(with BOM) for msvc]
```c
#include <stdio.h>
#include <locale.h>
#include <wchar.h>

int main(int argc, char *argv[]) {
	wchar_t *str = L"好久不见";

	setlocale(LC_ALL, "");

	wprintf(L"%ls\n", str);

	return 0;
}
```

R32:
To output Chinese characters correctly on the Console. 
close #53

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -21,6 +21,7 @@
  */
 #include <hl.h>
 #include <hlmodule.h>
+#include <locale.h>
 
 #ifdef HL_VCC
 #	include <crtdbg.h>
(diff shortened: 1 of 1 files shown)


ncannasse:
On Linux uprintf is implemented in ucs2.c by first converting to UTF-8, I guess this is not good either ?

R32:
Chinese Linux the default encoding is UTF-8. (Other Asian countries should be the same).
and i think `setlocale` does not affect the output of `printf` with `%s`

PR #57, opened 2017-03-20
Add controller rumble

pperidont:


ncannasse:
you forgot a haptic = null in close() ;)

ncannasse:
also API wise, I think you should not require rumbleInit, instead try to init (only once) when rumble() is called. You can keep a "isRumbleSupported" if you want it to be detected by user

pperidont:
It should be better now :)

PR #58, opened 2017-03-20
OpenAL Extensions

mcapdegelle:
Adds extension support for OpenAL.
- Loads all supported extensions by calling AL.loadExtensions & ALC.loadExtensions
- EFX related functions & constants (used for reverb in heaps) are packed in EFX.hx
- We might want to create a file per extenstion, but for now, all other extensions are packed into Soft.hx
- Mac specific extensions are not supported yet (like ASA for the reverb)
- Mac build uses openal-soft by default but can still use the standard openAL implementation by removing the "openal_soft" definition


Justinfront:
What advantages has the mac specific version of openAL, apple have a habbit of adjusting stuff or removing stuff without notice, my preference is to make the mac install as linux'y as possible so that it's more future proof/independant and so that the install process is more similar to linux one pooling user setup knowledge. Also in favour of making linux install less standard if it can share more of its approach with the Mac one.  User base for haxe is quite small when something breaks it's ideal that the same problems are seen on mac an linux, more specific ones are likely fixed slower as it requires an expert in both platforms.  But maybe there are good reasons for mac specific I know that macport and homebrew can easily mess up ( especially when having to update or with new osx versions ) so I am very much in favour of the 'kha' submodule approach so ideally HL should include submodules for all the c/c++ code used just compare Kha and OpenFL user setup experiances.
@mcapdegelle are the mac specifc extensions one that should work on linux or ones that are really worthwhile?  I don't know anything about the OpenAL just mentioning my experiance with setting up haxe complex projects historically, currently hxcpp is always a variable compared to js although it is always improving, it would be good if HL can be like js, always just work out of the box the same everywhere, dependancy on mac preinstalled openal to me sound like long term risky decision?

PR #71, opened 2017-08-10
Fix a bounds error in hl_read_strings

c0rner:
The check for buffer overrun in hl_read_strings could never trigger due to a minor bug.


c0rner:
You're welcome

ncannasse:
Thanks for spotting it, you're right the bounds check was not effective, but your PR was not working correctly because it didn't increment sdata as it should.

PR #72, opened 2017-09-15
SDL & DirectX additions

pperidont:
- SDL: Allowed joystick background events (when the window has lost focus)
- SDL: init_once now throws an exception on failure (with SDL_GetError message)
- DirectX: Added detect_keyboard_layout and win_clip_cursor
- DirectX: Added handling of window close event

### DIFF
--- a/libs/directx/dx/Driver.hx
+++ b/libs/directx/dx/Driver.hx
@@ -700,4 +700,13 @@ class Driver {
 	static function dxDebugPrint( str : hl.Bytes ) {
 	}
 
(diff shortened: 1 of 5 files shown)


ncannasse:
`ClientToScreen(wnd, (LPPOINT)& rect + 1);` this seems to write something on the C stack ???

pperidont:
We need to apply `ClientToScreen` on the two points that define the rect.

For clarity I changed it to:
```
ClientToScreen(wnd, (LPPOINT)& rect.left);
ClientToScreen(wnd, (LPPOINT)& rect.right);
```

PR #83, opened 2017-10-10
DirectX GameController

pperidont:
Adds support for both XInput and DirectInput.
DirectInput mappings are based on SDL and is mainly used for Dualshock4.
Requires `dinput8.dll` and `xinput9_1_0.dll` that are both parts of Windows (Vista and next versions).


ncannasse:
waiting for heaps hxd.Pad PR ;)

PR #96, opened 2017-11-22
Improved precision of GC_PROFILE timings

pperidont:


ncannasse:
Up : adding sys_precise_time (or using performance counter for sys_time ?) would be better

ncannasse:
(in all cases system code should be in sys.c, not alloc.c)

ncannasse:
up :) -- you can now use sys_time directly

src/hl.h
```diff
@@ -707,6 +707,7 @@ typedef struct {
 HL_API void *hl_fatal_error( const char *msg, const char *file, int line );
 HL_API void hl_fatal_fmt( const char *file, int line, const char *fmt, ...);
 HL_API void hl_sys_init(void **args, int nargs, void *hlfile);
+HL_API double hl_sys_time();
```
ncannasse: I prefer this to be declared as extern in alloc.c than made a public "official api" function in hl.h

Makefile
```diff
@@ -18,7 +18,7 @@ PCRE = include/pcre/pcre_chartables.o include/pcre/pcre_compile.o include/pcre/p
 	include/pcre/pcre_exec.o include/pcre/pcre_fullinfo.o include/pcre/pcre_globals.o \
 	include/pcre/pcre_newline.o include/pcre/pcre_string_utils.o include/pcre/pcre_tables.o include/pcre/pcre_xclass.o
 
-RUNTIME = src/alloc.o
+RUNTIME = src/alloc.o src/track.o
```
ncannasse: It this related to current PR?

PR #101, opened 2017-12-22
[WIP] Mobile support

rtissera:
This PR provides basic hashlink support for iOS, Android and tvOS platforms.


libs/sdl/sdl.c
```diff
@@ -80,9 +93,15 @@ HL_PRIM bool HL_NAME(init_once)() {
 	timeBeginPeriod(1);
 #	endif
 	// default GL parameters
+#if TARGET_OS_IOS || TARGET_OS_TV || __ANDROID__
```
ncannasse: please replace such checks with defined(HL_IOS) || defined(HL_ANDROID)
or maybe if there many of them you want to have  #ifdef HL_MOBILE ?

libs/sdl/sdl.c
```diff
@@ -142,6 +161,24 @@ HL_PRIM bool HL_NAME(event_loop)( event_data *event ) {
...
+				event->mouseX = e.tfinger.x*100;
+				event->mouseY = e.tfinger.y*100;
+				event->fingerId = e.tfinger.fingerId;
+				break;
+			case SDL_FINGERUP:
+				event->type = TouchUp;
+				event->mouseX = e.tfinger.x*100;
+				event->mouseY = e.tfinger.y*100;
+				event->fingerId = e.tfinger.fingerId;
+				break;
```
ncannasse: This block seems badly indented ;)

libs/sdl/sdl.c
```diff
@@ -427,6 +475,14 @@ HL_PRIM void HL_NAME(win_resize)(SDL_Window *win, int mode) {
...
 
 HL_PRIM void HL_NAME(win_swap_window)(SDL_Window *win) {
+#if TARGET_OS_IOS || TARGET_OS_TV
+	SDL_SysWMinfo info;
+	SDL_VERSION(&info.version);
+	SDL_GetWindowWMInfo(win, &info);
+
+	glBindFramebuffer(GL_FRAMEBUFFER, info.info.uikit.framebuffer);
+	glBindRenderbuffer(GL_RENDERBUFFER,info.info.uikit.colorbuffer);
+#endif
```
ncannasse: I think should be done at the Haxe GlDriver level, by exporting SDL_GetWindowWMInfo. But maybe it's a bit more complex to do, so if it's reliable enough we can leave it as it.

src/mobile/android/jni/hl_android.c
```diff
@@ -0,0 +1,80 @@
...
+	bool result;
+	const char* path = SDL_AndroidGetInternalStoragePath();
+	char* full_path = (char*)malloc(sizeof(char)*(strlen(path)+1+strlen(dir)+1));
+	strcpy(full_path, path);
+	strcat(full_path, "/");
+	strcat(full_path, dir);
+	return mkdir(full_path, mode);
+}
+
+#endif
```
ncannasse: This one is quite problematic, as we don't want to rely on SDL for hl standard library.
Also, they leak the mallocated memory.

Can't they be done using non SDL api calls (android c api) ?
If yes, we could have a single `sys_special_path` in src/std/sys.c

```C
const char *sys_special_path( const char*name ) {
#ifdef HL_ANDROID
    if( strcmp(path,"resource_path") == 0 ) return AndroidGetExternalStoragePath();
#endif
    return NULL;
}
```

Also, string concats should be done on Haxe side (this would also solve the malloc leak, and make mobile_create_directory useless)

src/mobile/ios/hl_ios.m
```diff
@@ -0,0 +1,110 @@
...
+    return [code UTF8String];
+}
+
+int hl_ios_get_retina_scale_factor()
+{
+	return [[UIScreen mainScreen] scale];
+}
+
+#endif
+#endif
```
ncannasse: Same remarks as for Android, hopefully these can be done using C API, or else we will have to add an extra sys_ios.m with these definitions (the retina_scale_factor should go into SDL, is there a SDL function to get the same result?)

src/std/file.c
```diff
@@ -44,6 +47,15 @@ HL_PRIM hl_fdesc *hl_file_open( vbyte *name, int mode, bool binary ) {
...
 	FILE *f = fopen((uchar*)name,MODES[mode|(binary?4:0)]);
+#   elif defined(HL_IOS) || defined(HL_TVOS) || defined(HL_ANDROID)
+    static const char *MODES[] = { "r", "w", "a", NULL, "rb", "wb", "ab" };
+    int m = mode | (binary?4:0);
+    FILE *f = NULL;
+    if (m==1 || m>=5){
+        f = fopen((char*)hl_mobile_get_document_path(name),MODES[m]);
+    }else{
+        f = fopen((char*)hl_mobile_get_resource_path(name),MODES[m]);
+    }
```
ncannasse: The prefixing of directories with a specific path should be done at Haxe level, this shouldn't be much a problem as Heaps as very few file reads

src/std/sys.c
```diff
@@ -282,15 +304,24 @@ HL_PRIM int hl_sys_command( vbyte *cmd ) {
...
+#endif
 	hl_blocking(false);
 	return WEXITSTATUS(status) | (WTERMSIG(status) << 8);
 #endif
 }
 
 HL_PRIM bool hl_sys_exists( vbyte *path ) {
+#if TARGET_OS_IOS || TARGET_OS_TV || __ANDROID__
+	return hl_mobile_file_exists(path);
+#else
```
ncannasse: Same remarks as for fopen : no prefixing at HL level

libs/sdl/sdl.c
rtissera: I think it would be really easier to leave it in the SDL C code right now. Yes, it is working fine on all iOS / tvOS devices.

src/std/sys.c
rtissera: This one and other similar cases have been fixed in last commit.

libs/sdl/sdl.c
rtissera: Yes, fixed in last commit with an added HL_MOBILE #define symbol.

libs/sdl/sdl.c
rtissera: Fixed in last commit.

src/hlc_main.c
```diff
@@ -20,6 +20,7 @@
  * DEALINGS IN THE SOFTWARE.
  */
 #include <hlc.h>
+#include <SDL_main.h>
```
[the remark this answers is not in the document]
rtissera: I wrapped this around some #ifdef HL_MOBILE as it is required to get SDL_Init working properly on mobile platforms.

src/mobile/android/jni/hl_android.c
rtissera: The SDL api calls are plain wrappers around Android JNI (C+Java) calls. 
I make use of these SDL api calls instead of plain JNI in order to keep the code clean and concise (rather than re-implementing the wheel). 

There is (at my best knowledge) no way to avoid the JNI C<=>Java bridge to achieve retrieving internal/external storage paths, except by using NativeActivity, which SDL2 is not using internally.

I do agree a sys_special_path function could be a good fit here for resolving file access specific issues on mobile, and we can re-implement a pure JNI AndroidGetExternalStoragePath() solution.

If that's the way you want to go let me know and I will move on this approach.

ncannasse: Yes, thanks for the details. Having sys_special_path using JNI API seems like the best way to me. This way it can be extended for other things as well later if necessary (including some paths not supported by SDL API).

libs/sdl/gl.c
```diff
@@ -381,7 +392,14 @@ HL_PRIM vdynamic *HL_NAME(gl_create_framebuffer)() {
 
 HL_PRIM void HL_NAME(gl_bind_framebuffer)( int target, vdynamic *f ) {
 	GLOG("%d,%d",target,ZIDX(f));
+#if	defined(HL_IOS) || defined(HL_TVOS)
+	SDL_SysWMinfo info;
+	SDL_VERSION(&info.version);
+	SDL_GetWindowWMInfo(SDL_GL_GetCurrentWindow(), &info);
```
ncannasse: Is the value returned constant ? if yes, could this be done + stored in a static local var so the calls are not made for every framebuffer change ? (or at least if ZIDX(f) == 0 in the worst case)

libs/sdl/gl.c
```diff
@@ -421,7 +439,14 @@ HL_PRIM vdynamic *HL_NAME(gl_create_renderbuffer)() {
 
 HL_PRIM void HL_NAME(gl_bind_renderbuffer)( int target, vdynamic *r ) {
 	GLOG("%d,%d",target,ZIDX(r));
+#if	defined(HL_IOS) || defined(HL_TVOS)
+	SDL_SysWMinfo info;
+	SDL_VERSION(&info.version);
+	SDL_GetWindowWMInfo(SDL_GL_GetCurrentWindow(), &info);
+	glBindRenderbuffer(GL_RENDERBUFFER, ZIDX(r) == 0 ? info.info.uikit.colorbuffer : ZIDX(r));
```
ncannasse: Same remark as above, let's try to reduce SDL calls.

libs/sdl/sdl.c
```diff
@@ -337,7 +378,11 @@ HL_PRIM SDL_Window *HL_NAME(win_create)(int width, int height) {
 }
 
 HL_PRIM SDL_GLContext HL_NAME(win_get_glcontext)(SDL_Window *win) {
-	return SDL_GL_CreateContext(win);
+	SDL_GLContext ctx = SDL_GL_GetCurrentContext();
+	if (ctx==NULL)
+		ctx = SDL_GL_CreateContext(win);
+	int result = SDL_GL_MakeCurrent(win ,ctx);
+	return ctx;
```
ncannasse: I think this is slightly problematic in case we want to have multiple Windows context. We should instead bind SDL_GL_GetCurrentContext and deal with it at the Haxe level (or HL_MOBILE if there's always a single context available?)

libs/sdl/sdl.c
ncannasse: Is that not a duplicate of the ZIDX()==0 checks ?

src/hlc_main.c
```diff
@@ -21,6 +21,10 @@
  */
 #include <hlc.h>
 
+#ifdef HL_MOBILE
+#   include <SDL_main.h>
+#endif
+
```
ncannasse: This one still needs to go away, what is it for actually?

rtissera: By including SDL_main.h we wrap the main() entry point with SDL code handling glue code for iOS app startup (app delegate).

We could skip including this header,  and use SDL_SetMainReady() function before SDL_Init() call into libs/sdl/sdl.c to circumvent SDL_Init() failing (it's just a stupid boolean check), but sadly it is not enough.

We would need to basically to copy/paste the SDL iOS AppDelegate code from SDL2 source code or rewrite a similar one to get a proper iOS app context. Another solution could be to copy/paste only SDL_uikitappdelegate.h from the SDL source tree, subclass it, and in hlc_main redefine main to SDL_main if we are on HL_MOBILE.

See https://github.com/spurious/SDL-mirror/blob/master/src/video/uikit/SDL_uikitappdelegate.m for more details on what the app delegate does.

Pretty similar situation on Android with SDL_main "magic" wrapping the SDLActivity Java class.

Feel free to share any better idea.

ncannasse: Ok, I think we can workaround this by simply checking `#if defined(HL_MOBILE) && defined(sdl__Sdl__val)` , this way this does not imply that HL_MOBILE always uses SDL

src/mobile/ios/hl_ios.m
ncannasse: Ok, but I want to avoid if possible the necessary to recompile Haxe code for a specific mobile platform (iOS or Android) unless the user does have some #if of its own. So that means we shouldn't have OS-specific hl_*** natives.

Maybe adding a single method in the following form would work:

```
#ifndef HL_MOBILE
const char *hl_sys_special( const char *key ) { 
     hl_error("Unknown sys_special key");
     return NULL;
}
#endif
```

So on traditional HL it will just return error with unknown special name (the above code)
On Android (src/std/sys_android.c) it would use JNI to return the paths
On iOS (src/std/sys_ios.m) it would allow access to both paths and the retina scale factor (as a string, but it can be parsed back on Haxe side).
Would that work for you ?

src/hlc_main.c
rtissera: Okay, just fixed that in latest commit.

src/mobile/ios/hl_ios.m
rtissera: Ok I will move on implementing it this way, sounds like a working plan.

libs/sdl/gl.c
rtissera: Fixed in https://github.com/HaxeFoundation/hashlink/pull/101/commits/1af597801d7ba76741dd36435f13b33acf9bbe16

libs/sdl/gl.c
rtissera: Fixed in https://github.com/HaxeFoundation/hashlink/pull/101/commits/1af597801d7ba76741dd36435f13b33acf9bbe16

libs/sdl/sdl.c
rtissera: Fixed in c324c0997fb26089fb1ce749e9ca86643e27467d

libs/sdl/sdl.c
```diff
@@ -80,9 +90,15 @@ HL_PRIM bool HL_NAME(init_once)() {
 	timeBeginPeriod(1);
 #	endif
 	// default GL parameters
+#ifdef HL_MOBILE
+    SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_ES);
+    SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 3);
+    SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 0);
+#else
```
ncannasse: There's a few places like this that uses 4 spaces instead of one tab, can you review?

src/std/process.c
```diff
@@ -273,6 +282,8 @@ HL_PRIM void hl_process_close( vprocess *p ) {
 HL_PRIM void hl_process_kill( vprocess *p ) {
 #	ifdef HL_WIN
 	TerminateProcess(p->pinf.hProcess,0xCDCDCDCD);
+#   elif defined(HL_IOS) || defined(HL_TVOS)
+    hl_error("hl_process_kill() not available on this platform");
```
ncannasse: space/tabs

src/std/sys.c
```diff
@@ -113,8 +113,14 @@ HL_PRIM vbyte *hl_sys_string() {
 	return (vbyte*)USTR("Mac");
 #elif defined(HL_CONSOLE)
 	return (vbyte*)sys_platform_name();
+#elif defined(HL_IOS)
+    return (vbyte*)USTR("iOS");
+#elif defined(HL_TVOS)
+    return (vbyte*)USTR("tvOS");
+#elif defined(HL_ANDROID)
+    return (vbyte*)USTR("Android");
```
ncannasse: space/tabs

src/std/sys.c
```diff
@@ -130,7 +136,11 @@ HL_PRIM vbyte *hl_sys_locale() {
 
 HL_PRIM void hl_sys_print( vbyte *msg ) {
 	hl_blocking(true);
+#ifdef HL_ANDROID
+	LOG_ANDROID(hl_to_utf8(msg));
+#else
```
ncannasse: maybe nitpicking, but would it not be the best to handle this directly in uprintf ?

src/std/sys_ios.m
```diff
@@ -0,0 +1,70 @@
...
+	return [[UIScreen mainScreen] scale];
+}
+
+const char *hl_sys_special( const char *key ) { 
+	if (strcmp(key, "ios_resource_path")==0)
+		return ios_get_resource_path();
+	else if (strcmp(key, "ios_document_path")==0)
+		return ios_get_document_path();
+	else if (strcmp(key, "ios_retina_scale_factor")==0)
+		return ios_get_retina_scale_factor();
```
ncannasse: I think you want to convert it to const char* here, and how are NSString to const char * garbage collected ? no risk of leak here ?

src/std/sys.c
rtissera: We need to use __android_log_print to get standard output through adb.

ncannasse: Sure, but I was thinking that could be handled at a slightly lower level (uprintf implementation) instead of hl_sys_print

rtissera: OK, no problem moving that in src/std/ucs2.c uprintf implementation.

src/std/sys_ios.m
rtissera: No object is owned or manually retained so it should not leak.
Howevefr I switched to a better model in latest commit making sure the returned string is not tied to the Objective-C runtime memory management.

ncannasse:
Seems like the Makefile was commited by error ?

src/hl.h
```diff
@@ -34,14 +34,37 @@
...
+#	define HL_ANDROID
+#	include <android/log.h>
+#	ifndef HL_ANDROID_LOG_TAG
+#		define HL_ANDROID_LOG_TAG "hl"
+#	endif
+#	ifndef HL_ANDROID_LOG_LEVEL
+#		define HL_ANDROID_LOG_LEVEL ANDROID_LOG_DEBUG
+#	endif
+#	define LOG_ANDROID(cfmt,cstr) __android_log_print(HL_ANDROID_LOG_LEVEL, HL_ANDROID_LOG_TAG, cfmt, cstr);
+#endif
```
ncannasse: Can we move the whole log support into ucs2.c since only it uses LOG_ANDROID now?

ncannasse:
@andyli could you make sure that src/std/sys_android.c is only included on android and src/std/sys_ios.m only included on tvOS in CMakefile ?

andyli:
Done in 5cf85c0d1ce9a04041b894bd8ad6c4a556a3e91e

nadako:
Can we have a sample android project somewhere that gives an idea on how to use this? :)

src/std/sys_android.c
```diff
@@ -0,0 +1,242 @@
...
+}
+
+/* JNI_OnLoad is automatically called when loading shared library through System.loadLibrary() Java call */
+JNIEXPORT jint JNICALL JNI_OnLoad(JavaVM* vm, void* reserved)
+{
+	JNIEnv *env;
+	jclass cls;
+
+	hl_java_vm = vm;
+	if ((*hl_java_vm)->GetEnv(hl_java_vm, (void**) &env, JNI_VERSION_1_4) != JNI_OK) {
```
nadako: @rtissera why this is needed if we overwrite `env` anyway later in `hl_android_jni_get_env` call?

PS I'm just asking because I'm new to all this, so trying to wrap my head around the initialization procedure.

src/std/sys_android.c
```diff
@@ -0,0 +1,242 @@
...
+
+#ifndef HL_JNI_LOG_TAG
+#define HL_JNI_LOG_TAG "HL_JNI"
+#endif
+
+#define HL_ANDROID_EXTERNAL_STORAGE_NOT_MOUNTED	(0)
+#define HL_ANDROID_EXTERNAL_STORAGE_MOUNTED_RO	(1)
+#define HL_ANDROID_EXTERNAL_STORAGE_MOUNTED_RW	(2 | HL_ANDROID_EXTERNAL_STORAGE_MOUNTED_RO)
+
+/* #define LOGI(...)  __android_log_print(ANDROID_LOG_INFO,HL_JNI_LOG_TAG,__VA_ARGS__) */
```
nadako: Any reason why these are commented out?

src/std/sys_android.c
```diff
@@ -0,0 +1,242 @@
...
+
+	hl_java_vm = vm;
+	if ((*hl_java_vm)->GetEnv(hl_java_vm, (void**) &env, JNI_VERSION_1_4) != JNI_OK) {
+		__android_log_print(ANDROID_LOG_ERROR, HL_JNI_LOG_TAG, "Failed to get the environment using GetEnv()");
+		return -1;
+	}
+
+	/* Create pthread "destructor" pthread key to detach properly all threads */
+	if (pthread_key_create(&hl_java_thread_key, hl_android_jni_thread_destructor) != 0) {
+		__android_log_print(ANDROID_LOG_ERROR, HL_JNI_LOG_TAG, "Error initializing pthread key");
```
nadako: Why not LOGE here and in other places?

src/hlc_main.c
```diff
@@ -21,6 +21,10 @@
  */
 #include <hlc.h>
 
+#if defined(HL_MOBILE) && defined(sdl__Sdl__val)
```
qkdreyer: What's the point of having **sdl__Sdl__val** defined here ?

ncannasse:
We need a way to call SdlMain() if we are compiling with SDL with HL/C

Le sam. 8 févr. 2020 à 22:08, Quentin Dreyer <notifications@github.com> a
écrit :

src/hlc_main.c
ncannasse: It's to make sure that SDL is included before the hl.h, because we might eventually want to compile for mobile without SDL framework.

PR #108, opened 2018-02-01
fix android startup crash because of the FindClass typo

nadako:
with this fix I actually was able to output a hello world with HL/C!

### DIFF
--- a/src/std/sys_android.c
+++ b/src/std/sys_android.c
@@ -101,7 +101,7 @@ JNIEXPORT jint JNICALL JNI_OnLoad(JavaVM* vm, void* reserved)
 	env = hl_android_jni_get_env();
 
 	/* Try to retrieve local reference to our Activity class */
-	cls = (*env)->FindClass(env, hl_java_activity_class);
+	cls = (*env)->FindClass(env, HL_ANDROID_ACTIVITY);
 	if (!cls) {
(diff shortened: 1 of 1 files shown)


nadako:
For those who interested, the `JNI_OnLoad` function that is called when you do `System.loadLibrary("hl")` impose some requirements:

* The activity class name should be whatever is defined with `HL_ANDROID_ACTIVITY` (by default: `org/haxe/HashLinkActivity`)
* The activity class must have a `public static Context getContext()` method that returns the application context.

Here's what worked for me for a simple HL/C hello world:

```java
package com.example.native_activity;

import android.app.Activity;
import android.content.Context;
import android.os.Bundle;

public class MainActivity extends Activity {
    private static MainActivity instance;

    @Override
    protected void onCreate(Bundle state) {
        super.onCreate(state);
        instance = this;

        System.loadLibrary("hl");
        System.loadLibrary("native-activity");

        startHL();
    }

    public static Context getContext() {
        return instance.getApplicationContext();
    }

    public native void startHL(); // see comment below
}
```

the `startHL` function is just something like this:

```c
JNIEXPORT jstring JNICALL Java_com_example_native_1activity_MainActivity_startHL(JNIEnv* env, jobject thiz) {
	main(0, NULL); // this is the `main` function from `hlc_main.c`
	               // or it could be custom initialization code that calls `hl_entry_point`
}
```

ncannasse:
@rtissera is it good to merge?

rtissera:
@ncannasse yes, my bad... thank you @nadako for the fix !

PR #111, opened 2018-02-05
Adding `extern "C"` to HL_PRIM if using CPP

starburst997:
While trying to compile hlc with directx I was having a bunch of linkage errors `unresolved external symbol `, adding `extern "C"` in front of the function in directx.cpp did the trick.

There might be a better way, if so I'd love to know it!

### DIFF
--- a/libs/directx/directx.cpp
+++ b/libs/directx/directx.cpp
@@ -44,7 +44,7 @@ static void ReportDxError( HRESULT err, int line ) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


ncannasse:
Please instead use a single  extern "C" { ... } block, or maybe we should add this automatically in HL_PRIM via hl.h #ifdef __cplusplus ?

starburst997:
Makes sense to add it to HL_PRIM, directx might not be the only case where a cpp file is needed. Pushed a new commit that do that.

starburst997:
I guess it might be needed in that `ifdef LIBHL_EXPORTS` block as well? Wasn't quite sure so I didn't.

PR #148, opened 2018-05-14
Missing (apparently required) null checks, on_alloc was insta-crashing on Linux

jcward:
Is somehow the first hash calc resulting in 0, matching the init conditions of prev_hash / prev_hash2 ?


ncannasse:
Oh, interesting I didn't hit into that. might be because hashing the empty string comes first. strange but still a valid fix.

PR #209, opened 2018-12-11
Add scanCode to SDL and DX keydown/keyup events.

Yanrishatum:
Introduces `Event.scanCode` to both SDL and DirectX on `KeyDown` and `KeyUp` events. 
Reasoning behind that is that `keyCode` acts more like `charCode` and depends on what active keyboard layout user have, making it less consistent.
And since we're not living in PS/2 era and all normal people use USB keyboard, scan codes should be persistent and reliable way that is not reliant on active layout. 

Related: HeapsIO/heaps#492


shakesoda:
@ncannasse can this be merged? it's useful.

PR #211, opened 2018-12-30
Add resizability configuration and screen centering for DirectX and SDL windows

RobertBorghese:
I wanted to provide some additional options to my own HashLink window in Heaps, so I went ahead and added a couple extra things to both window implementations. Figured they may be worth a pull request here.

_**Original Synopsis**_
The option to enable user resizing of the window can be enabled or disabled upon initialization (through constructor parameters) for both DirectX and SDL windows. A new method was created for the C bindings in order to maintain compatibility.

SDL was provided with position getters and a position setter method to match DirectX's implementation. Furthermore, both window implementations were provided with a method to center the window to the screen.

### DIFF
--- a/libs/directx/dx/Window.hx
+++ b/libs/directx/dx/Window.hx
@@ -18,9 +18,10 @@ class Window {
 
 	static var windows : Array<Window> = [];
 
-	public static inline var CW_USEDEFAULT : Int = 0x80000000;
+	public static inline var CW_USEDEFAULT = 0x80000000;
 
-	public static inline var RESIZABLE : Int = 0x000001;
+	public static inline var HIDDEN    = 0x000001;
+	public static inline var RESIZABLE = 0x000002;
 
 	var win : WinPtr;
 	var savedSize : { x : Int, y : Int, width : Int, height : Int };
--- a/libs/directx/dx/Window.hx
+++ b/libs/directx/dx/Window.hx
@@ -29,8 +29,8 @@ class Window {
 	public var visible(default, set) : Bool = true;
 	public var vsync : Bool;
 
-	public function new( title : String, width : Int, height : Int ) {
-		win = winCreate(width, height);
+	public function new( title : String, width : Int, height : Int, resizable : Bool = true ) {
+		win = winCreateEx(width, height, resizable);
 		this.title = title;
 		windows.push(this);
 		vsync = true;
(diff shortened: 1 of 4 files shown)


ncannasse:
Thanks! Here's a few comments:

DirectX: we should have a "Int" flags instead of "bool resizable", with bit 1 for resizable, which allows other flags later

SDL : as the wrapper try to be a 1:1 matching with SDL, we should expose the SDL window flags to the Haxe level, as well as SDL_WINDOWPOS_CENTERED and SDL_WINDOWPOS_UNDEFINED constants, and maybe also add x/y to winCreateEx

RobertBorghese:
Ah, thanks for the feedback! I'll rework to fit with those prerequisites and get back to you soon. 👍

RobertBorghese:
How does this look? I've gone ahead and exposed the SDL_WINDOW_* flags which can now be passed to the SDL Window constructor. While the OPENGL flag is mandatory, both SHOWN and RESIZABLE are now OR'd together as the default flag parameter.

The SDL_WINDOWPOS_* flags are now also exposed, and work with both the constructor and setPosition method.

On the DirectX side of things, the resizable boolean has now been replaced with a flag system as desired.  I've gone ahead and provided two flags: one for allowing resizing (RESIZABLE) and one for hiding upon initialization (HIDDEN). Similar to SDL, the x/y can now be placed in the constructor. In order to maintain behavior, the CW_USEDEFAULT value was also exposed.

Finally, in order to remedy a small bug in the original request, the window "style" is saved into the window's dx_events struct. That way it can be retrieved and reused when returning from fullscreen and other modes. Since the dx_events struct seems exclusively for event related behavior, I was somewhat torn whether the style should be stored there or with GetProp/SetProp. Perhaps the struct could be renamed/re-purposed for storing all window data? (A feature for setting a window's minimum/maximum size would also need to rely upon storing window-unique data.)

ncannasse:
Is it me or is SDL winCenter implementation is missing?

RobertBorghese:
Ah, I didn't think it would be needed anymore since the front-end "center" method no longer needed it. Figured it may be preferred to keep implementations on the Haxe-side of things when possible, perhaps not?

ncannasse:
Nevermind I was the one which mistake dx and sdl :)

PR #214, opened 2018-12-31
Add MP3 decoder to fmt

Yanrishatum:
Introduces MP3 decoder support. Utilizes [lieft/minimp3](https://github.com/lieff/minimp3) library to decode mp3.
Exposes two functions:
- `mp3_open(bytes, size)` - allocates reader structure.
- `mp3_decode_frame(mp3, pos, output, size, offset)` - reads frame at `pos` and outputs to `output + offset`.

Notes/concerns/details:
* I use `Float32` output mode instead of `Short`.
* I don't like double-buffering instead of direct write to `output` on `mp3_decode_frame`.
* There is no checks if frame will fit into output and it may corrupt memory due to that.
* Probably worth adding ability to read multiple frames at once, since 44100 mp3 will read a bunch of frames at once to fill the audio buffer.

I'm not very confident with C code, and would appreciate comments on how to improve this implementation.
Related: Heaps MP3 support

### DIFF
--- a/libs/fmt/fmt.c
+++ b/libs/fmt/fmt.c
@@ -11,6 +11,10 @@ extern bool sys_jpg_decode( vbyte *data, int dataLen, vbyte *out, int width, int
 #include <zlib.h>
 #include <vorbis/vorbisfile.h>
 
+#define MINIMP3_IMPLEMENTATION
+#define MINIMP3_FLOAT_OUTPUT
+#include "minimp3.h"
+
 /* ------------------------------------------------- IMG --------------------------------------------------- */
 
 typedef struct {
@@ -303,7 +307,7 @@ DEFINE_PRIM(_VOID, zip_flush_mode, _ZIP _I32);
 DEFINE_PRIM(_BOOL, inflate_buffer, _ZIP _BYTES _I32 _I32 _BYTES _I32 _I32 _REF(_I32) _REF(_I32));
 DEFINE_PRIM(_BOOL, deflate_buffer, _ZIP _BYTES _I32 _I32 _BYTES _I32 _I32 _REF(_I32) _REF(_I32));
 
-/* ------------------------------------------------- SOUND --------------------------------------------------- */
+/* ----------------------------------------------- SOUND : OGG ------------------------------------------------ */
 
 typedef struct _fmt_ogg fmt_ogg;
 struct _fmt_ogg {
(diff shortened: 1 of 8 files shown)


ncannasse:
For the same reason as https://github.com/HeapsIO/heaps/pull/514, I would prefer this to be part of a standalone library.

libs/fmt/fmt.c
```diff
@@ -418,6 +422,41 @@ DEFINE_PRIM(_I32, ogg_tell, _OGG);
...
+	return o;
+}
+
+HL_PRIM int HL_NAME(mp3_decode_frame)( fmt_mp3 *o, int position, char *output, int size, int offset ) {
+	int ret = -1;
+	hl_blocking(true);
+	ret = mp3dec_decode_frame(&o->dec, o->bytes + position, o->size - position, o->pcm, &o->info);
+	// TODO: Without memcpy?
+	// TODO: Safe write
+	memcpy( (void *)(output + offset), (void *)o->pcm, sizeof(o->pcm) );
```
ncannasse: I don't think sizeof(o->pcm) is the best here. You should only copy the amount of frame_bytes as specified in documentation. Please also handle errors (ret = 0) and return -1 without copying if the size to copy exceed `size`

ncannasse:
Looking at the minimal dependencies, I agree we can merge this without much concern, independently from the discussion in Heaps

libs/fmt/fmt.c
Yanrishatum: `dec.frame_bytes` refer to amount of bytes read from input buffer, e.g. assumed next frame offset. To copy only the amount of decoded samples (`pcm` uses 2-channel maximum size) it would need to calculate them trough `ret * o->info.channels * 4`.
As of errors:
`-1` when `position` is out of bounds of file bounds and when output buffer is too small.
`0` when no mp3 data was decoded.

Yanrishatum: Then output size should be as big as pcm, because it's impossible to know the sample count in frame before decoding it (or doing extra work on finding and reading header and interpreting it to detect amount of samples)

libs/fmt/fmt.c
```diff
@@ -418,6 +422,41 @@ DEFINE_PRIM(_I32, ogg_tell, _OGG);
...
+struct _fmt_mp3 {
+	char *bytes;
+	int size;
+	mp3dec_t dec;
+	mp3dec_frame_info_t info;
+	mp3d_sample_t pcm[MINIMP3_MAX_SAMPLES_PER_FRAME];
+};
+
+HL_PRIM fmt_mp3 *HL_NAME(mp3_open)( char *bytes, int size ) {
+	fmt_mp3 *o = (fmt_mp3*)hl_gc_alloc_raw(sizeof(fmt_mp3));
```
ncannasse: You should not use allow_raw here, as this will make the GC scan the whole data for pointers, eventually creating false positive. What you want here is hl_gc_alloc_noptr, or alloc_finalizer if some memory needs to be freed when the decoder is no longer in-use.

Yanrishatum: I'm not sure why, but if I use `alloc_noptr` or `alloc_finalizer` it throws `access_violation` when attempting to decode a frame.

ncannasse:
Both does not keep pointers so you have to keep a reference to bytes on
Haxe side so they don't get GCed.

Another option is to change the api so bytes gets passed for every decode
call.

Le sam. 5 janv. 2019 20:19, Pavel Alexandrov <notifications@github.com> a
écrit :

Yanrishatum:
Changed API to require passing bytes each time (Which also allows for livestreaming of mp3 as it downloads from internet), added error handling, fixed gc_alloc and also added `mp3_frame_info` helper function.

PR #218, opened 2019-01-13
Fix HL_IOS define

luboslenco:
Hello,

In order to compile for iOS I needed to slightly change the defines. I am not sure how valid this fix is. Maybe the `HL_IOS` define got outdated? Maybe I am missing something in the build process?

Compiling with Xcode Version 10.1 (10B61). Original build error was:
`/hl/src/hl.h:236:18: Cannot combine with previous 'type-name' declaration specifier`


ncannasse:
@rtissera any insight?

ncannasse:
@qkdreyer does your merged PR also address this?

qkdreyer:
> @qkdreyer does your merged PR also address this?

I believe this is no longer necessary since I was able to compile for iOS without it !

ncannasse:
Thanks!

PR #219, opened 2019-01-15
Fixed DirectX Cursor

RobertBorghese:
At the current moment, the DirectX's cursor implementation is completely broken. The cursor only changes for a bit, but reverts back immediately since the WM_SETCURSOR event is not handled. Another issue is that "hiding" the cursor doesn't just hide it in the client area, but the entire application including the title bar (making it difficult to move, close, etc.).

This pull request fixes both issues and also adds a method for checking cursor visibility on both DX and SDL.


ncannasse:
Thanks for the contribution

PR #223, opened 2019-01-19
Update libuv to v. 1.25 and add linux classes

flashultra:
Update to the latest libuv ( v. 1.25 ) and add all classes ( including linux ones)


ncannasse:
On linux etc, we prefer to use the system-installed library. The one made available in `include` is mostly provided to accomodate windows users.

PR #232, opened 2019-02-11
Fix macos warnings

dpeek:
Fix for deprecated `syscall(SYS_getid)` comes from here:
https://github.com/PowerShell/PowerShell/pull/2675/files

Is there some test I can do to verify this is working correctly? Also suspect return type (uint64) is an issue – bear in mind I know nothing about C programming, just trying to help ;)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ else ifeq ($(UNAME),Darwin)
 
 # Mac
 LIBEXT=dylib
-CFLAGS += -m$(ARCH) -I /opt/libjpeg-turbo/include -I /usr/local/opt/jpeg-turbo/include -I /usr/local/include -I /usr/local/opt/libvorbis/include -I /usr/local/opt/openal-soft/include -Dopenal_soft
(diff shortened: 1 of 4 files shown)


ncannasse:
hl_thread_id is only used in alloc.c to mark per-thread variables with the tid so the debugger knows which variables belongs to which thread. I doubt there will be some conflict there, and we don't yet have support for debugger on OSX so as long as the program links and runs I think we're good.

dpeek:
Heh, build and links on my machine? How do I see AppVeyor output?

ncannasse:
Details > click on the failing ENV

ncannasse:
@andyli not sure what's wrong with appveyor on Windows here, as we're only doing OSX changes

dpeek:
I don't _think_ that was me...

andyli:
It was because I used an undocumented way to call cmake and that was removed in recent cmake release.

https://cmake.org/cmake/help/v3.13/release/3.13.html#id2

> CMake now checks that at least one of the source or binary directory is specified when running CMake and issues an error if both are missing. This has always been a documented requirement, but the implementation previously accidentally accepted cases in which neither are specified so long as some other argument is given, and silently used the current working directory as the source and build tree.

Just fixed in master.

dpeek:
Rebased on master.

PR #236, opened 2019-03-14
Let macOS build set the SDK version

jgranick:
This change allows for the developer to specify a macOS SDK version. This is important in order to build HashLink for older macOS systems, while building on a newer system.

I debated whether this should have a different variable name (such as `SDK_VERSION`), but thought SDK would be easier to remember, and more convenient if you do not have to try and find the full SDK path yourself

```
make SDK=10.14
```


Makefile
```diff
@@ -64,6 +64,13 @@ else ifeq ($(UNAME),Darwin)
 LIBEXT=dylib
 CFLAGS += -m$(ARCH) -I /opt/libjpeg-turbo/include -I /usr/local/opt/jpeg-turbo/include -I /usr/local/include -I /usr/local/opt/libvorbis/include -I /usr/local/opt/openal-soft/include -Dopenal_soft  -DGL_SILENCE_DEPRECATION
 LFLAGS += -Wl,-export_dynamic -L/usr/local/lib
+
+ifneq ($(origin SDK), undefined)
```
ncannasse: can you instead use  ifdef with variable OSX_SDK ?

jgranick: Okay, updated

PR #240, opened 2019-04-04
use the actual module handle instead of the calling process handle (windows)

nadako:
As we discussed yesterday on Slack.

### DIFF
--- a/src/module.c
+++ b/src/module.c
@@ -24,7 +24,8 @@
 
 #ifdef HL_WIN
 #	include <windows.h>
-#	define dlopen(l,p)		(void*)( (l) ? LoadLibraryA(l) : GetModuleHandle(NULL))
+EXTERN_C IMAGE_DOS_HEADER __ImageBase;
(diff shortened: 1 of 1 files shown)


nadako:
That would mean reworking `hl_module_init_natives` though, which I would like not to do right now. I think it's fair to specify `builtin` to be the module from where the vm is running?

ncannasse:
Not sure, you might have a custom runtime exe that loads libhl.dll, in which case your patch will only look into libhl

nadako:
Assuming that by `libhl.dll` you mean a custom dll with hl module loader and not hashlink's `libhl.dll`, I thought you said that it's a fair restriction:

```
nadako [23:03]
 well then for windows we also need to loop over all loaded modules?

nicolas [23:04]
 depends on your definition of "builtin" . I guess it's a good enough restriction
```

Or have I misunderstood?

With this patch, I'm using a custom DLL with some builtin functions and a hl bootstrap/module loading code that is loaded by Godot exe and this works.

ncannasse:
I'll merge for now, we might fix that later ;)

PR #242, opened 2019-04-07
Update copyright notices to 2019

Gama11:


ncannasse:
copyrights does not require to be updated and I don't like how this usually clobber the file history ;)

Gama11:
I just saw that the copyright on the command line still says `-2018` and thought that I might as well update the rest too. I think that one should be updated? Otherwise it looks a bit like the last release was in 2018 / it's not maintained as actively as it actuallay is.

Gama11:
https://github.com/HaxeFoundation/hashlink/pull/243 ;)

Gama11:
Btw, I think if you give a range of years, it _does_ imply that the headers will be updated yearly, perhaps not legally, but otherwise. If you don't plan on updating them, it would probably be better if they just said `Copyright (C)2015 Haxe Foundation`.

PR #244, opened 2019-04-09
Hide js/gamepad subsystem init latency in a thread

shakesoda:
On some Windows 10 systems (I've yet to find any specific cause), doing this is **extremely** slow.

I'm getting tired of waiting 5-10 extra seconds on my dev machine before getting a window and it's entirely unclear if this is an SDL bug or a Windows bug. One theory about this (from @rcgordon) is
that it's a timeout from a non-responsive device, in which case this is probably the right idea (or at least a sane one) anyway.

This trades the time at window creation with some delay before you start receiving gamepad inputs, which is much more tolerable. I haven't noticed any negative/unexpected side effects (I've been using this for some days already), and on most systems this will init fast enough to not notice any difference.

Also, don't initialize the audio subsystem, it's not used/exposed.

### DIFF
--- a/libs/sdl/sdl.c
+++ b/libs/sdl/sdl.c
@@ -81,12 +81,23 @@ typedef struct {
 	int fingerId;
 } event_data;
 
+// workaround for _extremely_ slow (5-10 seconds) gamepad init on some windows 10 systems.
+// it's not usually a big deal if gamepads don't accept input for a few seconds as long as
+// we've got the window up and doing things otherwise. on most systems this will be fast.
+static int init_gamepad(void* ptr) {
+	SDL_InitSubSystem(SDL_INIT_JOYSTICK);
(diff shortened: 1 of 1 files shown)


ncannasse:
Could we at least wait for like 1s that the threads executes so that it doesn't change anything for the cases the pad initialize immediately (which is true most of the time actually).

ncannasse:
Please resubmit with the fix I asked ;)

shakesoda:
i know this is closed, but if anyone was using it i have updated it for latest master. turns out it also happens on win7, not just 10 as originally thought. i didn't put in a delay because the delay would be longer than my entire program start time is, and on unaffected systems there's no perceptible difference (unless you were doing something like holding a button on startup, where it might miss the first frame?).

PR #246, opened 2019-04-25
Unicode fixes

Aurel300:
- ~~convert to UTF-16 before printing in `sys_print` (similar to the fix in hxcpp)~~
 - ~~`0xD7C0` was used instead of `0xD800` for high surrogates~~
 - [x] handle surrogates when calculating string length
 - [x] handle surrogates when converting to UTF-8
 - [x] `Sys.putEnv("TEST_VAR", "😉"); trace(Sys.getEnv("TEST_VAR"));` did not produce the correct result (UTF-16 to UTF-8 conversion was wrong)
 - [x] https://github.com/HaxeFoundation/haxe/issues/8204 ? cannot reproduce yet

### DIFF
--- a/src/std/string.c
+++ b/src/std/string.c
@@ -225,7 +225,7 @@ HL_PRIM vbyte *hl_utf16_to_utf8( vbyte *str, int len, int *size ) {
 			out[p++] = (vbyte)(0xC0|(v>>6));
 			out[p++] = (vbyte)(0x80|(v&63));
 		} else if( v >= 0xD800 && v <= 0xDFFF ) {
-			int k = ((((int)v - 0xD800) << 10) | (((int)*++c) - 0xDC00)) + 0x10000;
+			int k = ((((int)v - 0xD800) << 10) | (((int)*++c) - 0xDC00));
 			out[p++] = (vbyte)(0xF0|(k>>18));
 			out[p++] = (vbyte)(0x80 | ((k >> 12) & 63));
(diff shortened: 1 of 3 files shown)


ncannasse:
This should not be needed. Uprintf implementation is responsible for doing that.

Aurel300:
What platform are you testing on? Does `trace("😉");` work for you? On Ubuntu 16.04 without this fix it prints the surrogate bytes (so it just shows `??????` in the terminal).

ncannasse:
What I meant is that it should be fixed by modifying uprintf implementation instead of changing only hl_sys_print.

I'm testing on Windows. And yes it's using wprintf which does not seem to correctly convert to utf8. But this seems quite tricky because Windows terminal is not utf8 so trace("é") works with wprintf in cmd but does not work when converting to utf8, it shows Ã© in terminal.

ncannasse:
I fixed it for windows in 3c6164d88a2b812f3e6bfa1713c303e94beb7f1f (+ just added stderr as well)
@hughsando you might want to use that as well, it sets stdout to utf8 so wprintf actually works well when printing both to terminal and redirecting to a file

@Aurel300 on other platforms, look at uprintf implementation in ucs2.c

Aurel300:
Yes, you're right – I thought `uprintf` was a built-in. But `uprintf` doesn't actually handle surrogate conversion, so I'll fix that.

PR #252, opened 2019-05-28
added accelerometer events

josuigoa:


ncannasse:
That patch seems to be dangerous as there does not seem to be any guarantee that the axis index reported by the joystick API does not overlap with a game controller button.

This is what I understand from
https://discourse.libsdl.org/t/difference-between-joysticks-and-game-controllers/24028/2

josuigoa:
do you suggest to create a new event type (AccelerometerAxis, e.g.)?

ncannasse:
It would also require events for adding/removing joysticks, etc.

josuigoa:
I've implemented the joystick API. It is ok for you to mantain the ```SDL_SetHint(SDL_HINT_ACCELEROMETER_AS_JOYSTICK, "1");``` as is, or do you prefer to put it inside a conditional flag?

ncannasse:
I guess the best would be to give access to the SDL_SetHint method from Haxe.

PR #264, opened 2019-06-27
Fix duplicate LF->CRLF encoding

Aurel300:
See HaxeFoundation/haxe#8379

 - in `hl_sys_print`, switch stdout to binary mode, convert string to UTF-8, output it, then switch back
 - in `ucs2.c` make the functions `ustrlen_utf8`, `utostr`, and `utos` always be defined
 - remove alias `utostr` -> `wcstombs` on Windows, since it relies on the locale type


Aurel300:
Closed, see https://github.com/HaxeFoundation/haxe/issues/8379#issuecomment-507255656

PR #266, opened 2019-07-09
Add UTC-based methods

Aurel300:
See HaxeFoundation/haxe#8508

 - add `date_get_utc_inf` - same as `date_get_inf`, but using `gmtime_r` instead of `localtime_r`


ncannasse:
Is there gmtime_r on Windows ?

Aurel300:
> Is there gmtime_r on Windows ?

No, but there already is a wrapper defined for Windows for both `_r` functions:

https://github.com/HaxeFoundation/hashlink/blob/10518535b7ece99343fa3ece723160f4dcf28f8c/src/std/date.c#L39-L44

Issue #280, opened 2019-08-15
Sys.time()


ConstNW:
on windows Sys.time() return high precision timestamp but not in unix epoch as expected.

```haxe
class Main
{
    public static function main( ) : Void
    {
        trace(Sys.time());
        trace(Date.now().getTime());
        trace(haxe.Timer.stamp());
    }
}
```
got output:
```
src/Main.hx:23: 14583.3435118
src/Main.hx:24: 1565876592000
src/Main.hx:25: 14583.9046364
```
expected output :
```
src/Main.hx:23: 1565876154.966
src/Main.hx:24: 1565876154000
src/Main.hx:25: 1565876154.969
```

ConstNW:
updated version output:
```
src/Main.hx:23: 1565879082.43818
src/Main.hx:24: 1565879082000
src/Main.hx:25: 1565879082.43865
```

Aurel300:
I assume the comment from [`haxe.Timer.stamp`](https://api.haxe.org/v/development/haxe/Timer.html#stamp) applies here too?

> The value itself might differ depending on platforms, only differences between two values make sense.

Aurel300:
I understand, and your fix seems good. I was just trying to say that that comment should have been put on `Sys.time`. And I wonder if it is not epoch-based on any other targets.

ConstNW:
as i know every sys-target got unix-epoch from Sys.time except hashlink

PR #281, opened 2019-08-15
update hl_sys_time on Windows

ConstNW:
#280

### DIFF
--- a/src/std/sys.c
+++ b/src/std/sys.c
@@ -163,14 +163,18 @@ HL_PRIM void hl_sys_exit( int code ) {
 
 HL_PRIM double hl_sys_time() {
 #ifdef HL_WIN
+	#define EPOCH_DIFF	(134774*24*60*60.0)
 	static double freq = 0.;
 	LARGE_INTEGER time;
+	FILETIME ft;
 	if( freq == 0 ) {
 		QueryPerformanceFrequency(&time);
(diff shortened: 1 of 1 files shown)


src/std/sys.c
```diff
@@ -163,14 +163,18 @@ HL_PRIM void hl_sys_exit( int code ) {
 
 HL_PRIM double hl_sys_time() {
 #ifdef HL_WIN
+	#define EPOCH_DIFF	(134774*24*60*60.0)
```
Aurel300: What is this value based on?

ConstNW: it's epoch diff between Unix epoch and Windows FILETIME epoch (started January 1, 1601 (UTC))

ncannasse:
This new sys_time is 1.6 slower than previous implementation but I guess it's good enough.
Please note that there also might be some non linear timings, as explained here https://stackoverflow.com/questions/34557407/queryperformancecounter-or-getsystemtimepreciseasfiletime-when-using-sntp

Shoozza:
Btw "GetSystemTimePreciseAsFileTime" is only available on Windows 8+. So this patch removed support for earlier versions of Windows.

https://docs.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getsystemtimepreciseasfiletime

PR #285, opened 2019-08-21
update hl_sys_time to support OS below Windows 8

ConstNW:
#281 
_Originally posted by @Shoozza in https://github.com/HaxeFoundation/hashlink/pull/281#issuecomment-523223056_

[GetSystemTimePreciseAsFileTime](https://docs.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getsystemtimepreciseasfiletime) supported since Windows 8
[GetSystemTimeAsFileTime](https://docs.microsoft.com/ru-ru/windows/win32/api/sysinfoapi/nf-sysinfoapi-getsystemtimeasfiletime) supported since Windows 2000 Professional

### DIFF
--- a/src/std/sys.c
+++ b/src/std/sys.c
@@ -164,21 +164,30 @@ HL_PRIM void hl_sys_exit( int code ) {
 HL_PRIM double hl_sys_time() {
 #ifdef HL_WIN
 	#define EPOCH_DIFF	(134774*24*60*60.0)
+	static double time_diff = 0.;
 	static double freq = 0.;
 	LARGE_INTEGER time;
+	LARGE_INTEGER start_time;
 	FILETIME ft;
+
 	if( freq == 0 ) {
 		QueryPerformanceFrequency(&time);
 		freq = (double)time.QuadPart;
 	}
-	#if (_WIN32_WINNT < _WIN32_WINNT_WIN8)
-	GetSystemTimeAsFileTime(&ft);
-	#else
-	GetSystemTimePreciseAsFileTime(&ft);
-	#endif
-	time.LowPart = ft.dwLowDateTime;
-	time.HighPart = ft.dwHighDateTime;
-	return ((double)time.QuadPart) / freq - EPOCH_DIFF;
+	QueryPerformanceCounter(&time);
+
+	if( time_diff == 0 ) {
+		#if (_WIN32_WINNT < _WIN32_WINNT_WIN8)
+		GetSystemTimeAsFileTime(&ft);
+		#else
+		GetSystemTimePreciseAsFileTime(&ft);
+		#endif
+		start_time.LowPart = ft.dwLowDateTime;
+		start_time.HighPart = ft.dwHighDateTime;
+		time_diff = ((double)start_time.QuadPart - (double)time.QuadPart) / freq - EPOCH_DIFF;
+	}
+
+	return time_diff + ((double)time.QuadPart) / freq;
 #else
 	struct timeval tv;
 	if( gettimeofday(&tv,NULL) != 0 )
--- a/src/std/sys.c
+++ b/src/std/sys.c
@@ -177,11 +177,7 @@ HL_PRIM double hl_sys_time() {
 	QueryPerformanceCounter(&time);
 
 	if( time_diff == 0 ) {
-		#if (_WIN32_WINNT < _WIN32_WINNT_WIN8)
 		GetSystemTimeAsFileTime(&ft);
-		#else
-		GetSystemTimePreciseAsFileTime(&ft);
-		#endif
 		start_time.LowPart = ft.dwLowDateTime;
 		start_time.HighPart = ft.dwHighDateTime;
 		time_diff = ((double)start_time.QuadPart - (double)time.QuadPart) / freq - EPOCH_DIFF;
--- a/src/std/sys.c
+++ b/src/std/sys.c
@@ -171,7 +171,11 @@ HL_PRIM double hl_sys_time() {
 		QueryPerformanceFrequency(&time);
 		freq = (double)time.QuadPart;
 	}
+	#if (_WIN32_WINNT < _WIN32_WINNT_WIN8)
+	GetSystemTimeAsFileTime(&ft);
+	#else
 	GetSystemTimePreciseAsFileTime(&ft);
+	#endif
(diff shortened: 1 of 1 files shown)


Aurel300:
What is the precision of `GetSystemTimeAsFileTime`? Is it at least as precise as `QueryPerformanceCounter` (which is also available since Win2k)?

ConstNW:
According to that post  https://stackoverflow.com/questions/34557407/queryperformancecounter-or-getsystemtimepreciseasfiletime-when-using-sntp
```
| Function                       | Return type       | Resolution | Timezone |
|--------------------------------|-------------------|------------|----------|
| GetSystemTimeAsFileTime        | FILE_TIME (0.1us) | ~10-15 ms  | UTC      |
| GetSystemTimeAsPreciseFileTime | FILE_TIME (0.1us) | 0.1 us     | UTC      |
|                                |                   |            |          |
| QueryPerformanceCounter        | Int64 (ticks)     | 0.1 us     | n/a      |

```

ncannasse:
We don't want 10-15 Ms resolution for haxe.timer.stamp

ConstNW:
Look's like 'resolution' is made on wrong assumption.
in practice 'resolution' is about 0.1ms

Sample code:
```haxe
class Main
{
    public static function main( ) : Void
    {
        var time1 = Sys.time();
        var time2 = Sys.time2();
        for( i in 0...100 )
        {
            var t1 = Sys.time();
            var t2 = Sys.time2();
            if( t2 != time2 ) trace('new time ' + t2);
            trace(t1-t2);
            time1 = t1;
            time2 = t2;
        }
    }
}
```

Output:
```
src/Main.hx:11: new time 1566395753.58509
src/Main.hx:12: 1.33514404296875e-05
src/Main.hx:12: 6.67572021484375e-05
src/Main.hx:12: 9.1552734375e-05
src/Main.hx:12: 0.000116348266601563
src/Main.hx:12: 0.0001373291015625
src/Main.hx:12: 0.000162124633789063
src/Main.hx:12: 0.000185012817382813
src/Main.hx:12: 0.00020599365234375
src/Main.hx:12: 0.000230789184570313
src/Main.hx:12: 0.000253677368164063
src/Main.hx:12: 0.000276565551757813
src/Main.hx:12: 0.000299453735351563
src/Main.hx:12: 0.000324249267578125
src/Main.hx:12: 0.000345230102539063
src/Main.hx:12: 0.000368118286132813
src/Main.hx:12: 0.000392913818359375
src/Main.hx:12: 0.000413894653320313
src/Main.hx:12: 0.000438690185546875
src/Main.hx:12: 0.000461578369140625
src/Main.hx:12: 0.000484466552734375
src/Main.hx:12: 0.000507354736328125
src/Main.hx:12: 0.000530242919921875
src/Main.hx:12: 0.000553131103515625
src/Main.hx:12: 0.000576019287109375
src/Main.hx:12: 0.000600814819335938
src/Main.hx:12: 0.000621795654296875
src/Main.hx:12: 0.000644683837890625
src/Main.hx:12: 0.000669479370117188
src/Main.hx:12: 0.000692367553710938
src/Main.hx:12: 0.000713348388671875
src/Main.hx:12: 0.000738143920898438
src/Main.hx:12: 0.000761032104492188
src/Main.hx:12: 0.000782012939453125
src/Main.hx:12: 0.000806808471679688
src/Main.hx:12: 0.000829696655273438
src/Main.hx:12: 0.000850677490234375
src/Main.hx:12: 0.000875473022460938
src/Main.hx:12: 0.000898361206054688
src/Main.hx:12: 0.000919342041015625
src/Main.hx:12: 0.000944137573242188
src/Main.hx:12: 0.000967025756835938
src/Main.hx:11: new time 1566395753.58607
```

Aurel300:
If I'm reading that right the resolution is actually about 1ms (`0.000967025756835938` ~= `0.001`). But is this the behaviour on lower versions of Windows as well?

ConstNW:
yes, my wrong. Resolution is about 1ms
[The Windows Timestamp project](http://www.windowstimestamp.com/description) talk about ActualResolution of 0.9765625 ms

and yes, on lower versions of windows is same 1ms resolution.

shakesoda:
@ncannasse isn't this bad for timer.stamp anyway? it's not monotonic anymore since changing away from QPC.

ncannasse:
Yes, we should revert back to previous implementation, given we want both fast and precise measure.
@ConstNW could you submit a version were we use GetSystemAsFileTime only once at initialisation time, to calculate the offset vs QueryPerformanceCounter, then only use QPC afterwards with this offset - the way it used to be?

ConstNW:
Sure, it will be ok

ncannasse:
I'm still worried that these higher bits might reduce the double precision by losing the lower ones. This is something we should check as well for haxe.Timer.stamp

ncannasse:
@ConstNW Thanks ! Could you check that the delta between two consecutive not equal values is the same with or without the offset?

Also, could you only use GetSystemTimeAsFileTime so binaries compiled on Windows10 are also 7-compatible, and perform the init at the same time as freq?

ConstNW:
With offset
```
src/Main.hx:10: 2.14576721191406e-06
src/Main.hx:10: 0.000194549560546875
src/Main.hx:10: 2.88486480712891e-05
src/Main.hx:10: 2.76565551757813e-05
src/Main.hx:10: 2.47955322265625e-05
src/Main.hx:10: 2.45571136474609e-05
src/Main.hx:10: 2.43186950683594e-05
src/Main.hx:10: 2.43186950683594e-05
src/Main.hx:10: 2.59876251220703e-05
src/Main.hx:10: 2.38418579101563e-05
```

Without offset
```
src/Main.hx:10: 1.00000761449337e-07
src/Main.hx:10: 0.000224799994612113
src/Main.hx:10: 3.03999986499548e-05
src/Main.hx:10: 2.69000011030585e-05
src/Main.hx:10: 2.46000126935542e-05
src/Main.hx:10: 2.38000066019595e-05
src/Main.hx:10: 2.39999790210277e-05
src/Main.hx:10: 2.37000058405101e-05
src/Main.hx:10: 2.39000073634088e-05
src/Main.hx:10: 2.37000058405101e-05
```

ncannasse:
We're losing about 3% precision on the lower scores (worst case scenario), so I guess that's good enough

PR #292, opened 2019-09-02
add Lock bindings to thread.c

ConstNW:
porting nekovm code by @ncannasse  [suggestion](https://github.com/HaxeFoundation/haxe/issues/8699#issuecomment-525023553) at HaxeFoundation/haxe/issues/8699

### DIFF
--- a/src/std/thread.c
+++ b/src/std/thread.c
@@ -51,6 +51,8 @@ struct _hl_tls {
 #	include <pthread.h>
 #	include <unistd.h>
 #	include <sys/syscall.h>
+#	include <sys/time.h>
+
 
 struct _hl_mutex {
 	void (*free)( hl_mutex * );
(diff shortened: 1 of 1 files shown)


src/std/thread.c
```diff
@@ -338,6 +340,112 @@ DEFINE_PRIM(_VOID, deque_add, _DEQUE _DYN);
...
 
+// ----------------- LOCK
+
+struct _hl_lock;
+typedef struct _hl_lock hl_lock;
+
+struct _hl_lock {
+	void (*free)( hl_lock * );
+#if !defined(HL_THREADS)
+	void *value;
```
ncannasse: why having an unused value here?

ConstNW: it's useless.

Simn:
Can we merge this? Quick reply appreciated!

sonygod:
@Simn it's work!

PR #297, opened 2019-09-05
add PCRE_NO_UTF8_CHECK option

ConstNW:
Option to disable utf8 checks every regexp_match call

check off: `0.11178874969482s`
check on: `2.42371964454651s`

Sample code:
```haxe
class Main 
{
	static function main( )
	{
		var str = StringTools.lpad('', '"', 10 * 1024);
		var r = ~/"/g;
		haxe.Timer.measure(function(){
			for(i in 0...10)
				replace(r, str);
		});
	}

	static function replace( r : EReg, str : String )
	{
		str = r.replace(str, '');
	}
}
```


Simn:
Nice, just added this flag to the eval PCRE calls and the benchmarks are now 50x faster.

PR #320, opened 2019-10-23
Added error popup for failed SDL window creation

jefvel:
The dialog will show a bit more detailed info retrieved from `SDL_GetError`.


ncannasse:
I prefer if errors are handled on the Haxe side.

PR #323, opened 2019-11-08
Fix sqlite.hdll and mysql.hdll building on Linux

jonasmalacofilho:
Each of the build systems would build one of these, but not both.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ MARCH ?= $(LBITS)
 PREFIX ?= /usr/local
 INSTALL_DIR ?= $(PREFIX)
 
-LIBS=fmt sdl ssl openal ui uv mysql
+LIBS=fmt sdl ssl openal ui uv mysql sqlite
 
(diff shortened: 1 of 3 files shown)


ncannasse:
I think it should use the system installed libsqlite instead of the "include" one. Also it should follow fmt that has its own include paths instead of adding itself to CFLAGS

jonasmalacofilho:
> I think it should use the system installed libsqlite instead of the "include" one

Just following what was already in the repository and in use by cmake, but I'll gladly change this.

> Also it should follow fmt that has its own include paths instead of adding itself to CFLAGS

I initially thought and did so, except that failed to build in CI, making me notice that those targets are misleading.  The objects are not built with the explicit target, but rather through an implicit one.

In fact, the second commit changes the fmt target to remove those unnecessary `-I` paths and avoid further confusion.

jonasmalacofilho:
I'm going to close this PR because it's not really relevant or useful anymore... the problem still exists, but the better fix (see comments above) will look nothing like this, and I'm not working on this.

I also apologize for saying "I'll gladly change this" and seemingly never actually getting to it.  As I remember it, CMake was being annoying (or at least making me annoyed) and this got preempted by other more important or actionable things, and only then eventually forgotten at the bottom of my to-do stack.

PR #342, opened 2020-01-07
mimalloc

Aurel300:
This PR:

 - [x] adds mimalloc sources to `include` dir (as of https://github.com/microsoft/mimalloc/commit/acb03c54971c4b0a43a6d17ea55a9d5feb88972f)
 - [x] adds mimalloc to `cmake` and `make` build steps
 - [x] replaces usages of `malloc` and `free` with `mi_~` in `src/alloc.c`
 - [x] replaces `gc_alloc_page_memory` with `mi_malloc_aligned` (except on consoles)

Possible TODO:

 - replace all other `malloc` and `free` usages in the sources with `mi_~`

Also of note is that mimalloc depends on `stdatomic.h` which is a C11 feature, I think that bumps our mimimum required compiler version (at least it did for me).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+/include/** linguist-vendored=true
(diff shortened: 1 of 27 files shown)


ncannasse:
mimalloc has some issues (can't compile with old vs2013 for win7 support). Closing this for now.

PR #343, opened 2020-01-16
New gc

ncannasse:


R32:
Is it still a **mark-and-not-sweep** collector?

Aurel300:
@R32 yes, this was just a refactor without changing how the GC works. For now :)

Issue #164, opened 2018-06-27
Better management of big blocks


ncannasse:
Big blocks should be handled using a more traditional free list and reserve a significant amount of the memory space in 32 bits.

Also, we want to have a manual free() for big blocks so we don't wait for GC to run before releasing it for reusage (for instance loading many textures).

ncannasse:
We might want to use something like https://github.com/microsoft/mimalloc to handle block allocation within pages. Would still require to register the pages for fast check if pointer within page.

ncannasse:
Big blocks are now page aligned and page headers are outside of block data.

Issue #298, opened 2019-09-06
performance of Array construct to improve  


sonygod:
performance of Array construct  improve  

compare between Hl ,Neko ,Nodejs,hxcpp

hl:1000ms

Neko:2000ms

node:32ms

hxcpp:~=37.74ms

php:228ms

   	package;

	/**
	 * ...
	 * @author sonygod
	 */
	class ArrayTest 
	{

		static function main() 
		{
			var t1 = Date.now().getTime();
				for (i in 10...100000) {
				var arr:Array<String> = Std.string(i).split('');//var arr:Array<String> cost lot's of cpu.
				}
				
				trace(Date.now().getTime() - t1);
		}
		
	}

![image](https://user-images.githubusercontent.com/748320/64404897-03516c00-d0b0-11e9-87f7-12854fd01fdd.png)

ncannasse:
This is more likely GC related, not "Array constructor" (or another feature) related. It will be improved with better gc performances when we upgrade GC sometimes in the future. Please avoid in the meanwhile to post benchmarks that are mainly related to allocation.

sonygod:
@ConstNW  What do you think?

PR #344, opened 2020-01-25
Added $freelist to GC allocator

R32:
* Added `gc_freelist` for gc allocator which will be reset on every GC start
* Added `hl_bytes_free()` that maybe works for #164 , and it's **DANGEROUS** if you free a `PTR` twice or write something on the freed block.

I hope I didn't mess up something :)

### DIFF
--- a/src/allocator.c
+++ b/src/allocator.c
@@ -273,8 +273,7 @@ alloc_var:
 }
 
 static void* gc_freelist_pickup(int* allocated, int part, int kind) {
-	gc_pheader *page = gc_free_pages[(part << PAGE_KIND_BITS) | kind];
-	if (part >= GC_FIXED_PARTS && kind != MEM_KIND_FINALIZER && page && page->bmp) {
+	if (part >= GC_FIXED_PARTS && kind != MEM_KIND_FINALIZER) {
 		int nblocks = *allocated >> GC_SBITS[part];
 		int index = GC_FL_OFFSET(part, nblocks);
 		if (index < GC_FL_MAX) {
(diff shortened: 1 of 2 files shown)


sonygod:
+1

@ncannasse  ,will you take a look ?

R32:
Some notes:
* `gc_freelist will be reset at GC start, so there will be no fragments.
* The new `hl_bytes_free()` maybe works for #164 , and it's quite **DANGEROUS** if you free a `PTR` twice or write something on the freed block.

sonygod:
@R32 

it's seem nothing was improved when have test here

https://github.com/HaxeFoundation/hashlink/issues/298

ubuntu  18.04  4cores and 4 G


hl: use your version and cost more than 5000ms 

hxcpp:only take 256ms

sonygod:
@R32 

you last change got the same result as HLC ,

4000ms

it's seem got a little improve gc now.


    root@ets-game:~/test# cd testgc
    root@ets-game:~/test/testgc# vi ArrayTest.hx
    root@ets-game:~/test/testgc# haxe -cp ./ -main ArrayTest -hl test.hl
    root@ets-game:~/test/testgc# ls
    ArrayTest.hx  test.hl
    root@ets-game:~/test/testgc# hl test.hl
    ArrayTest.hx:15: 4000
    root@ets-game:~/test/testgc# haxe -cp ./ -main ArrayTest -hl out/main.c
    Code generated in out/main.c automatic native compilation not yet implemented
    root@ets-game:~/test/testgc# gcc -O3 -o myapp -std=c11 -I out out/main.c -lhl
    root@ets-game:~/test/testgc# ./myapp
    ArrayTest.hx:15: 4000

ncannasse:
I appreciate the effort to deal with GC but I don't think that I can merge as it because each change would be subject to discussion and pro/cons.

On my side I'm considering a few changes, but very incremental and fully tested with our games before merge:
- allocate dedicated pages for blocks larger than 1MB (not using "variable" pages)
- faster dispatch from size to alloc_var / fixed
- rethink entirely how we use marking bits for allocation - currently we need to rescan all pages bits which makes allocations slower right after a major gc.

@Aurel300 is also working on a entirely rewritten Immix GC as part of his thesis

R32:
I will do  `git rebase`, and then provide a more detailed comparison on pro/cons :)

PR #354, opened 2020-03-12
build: prevent fixing SDL2 header path for iOS

qkdreyer:
This PR allows me to prevent applying this git [patch](https://gist.github.com/qkdreyer/1a22b10fe4064554f724bf5a400fe470) in order to be able to compile hashlink for iOS using Xcode


qkdreyer:
@ncannasse https://github.com/microsoft/azure-pipelines-image-generation/blob/master/README.md

qkdreyer:
@ncannasse I've also successfully fixed azure pipelines

ncannasse:
Could you keep /usr/local/include for backward compatibility (just in case) ?

qkdreyer:
Done ✅

tobil4sk:
@qkdreyer do you remember why this patch was needed? The `#include <SDL.h>` lines don't work when building on macos systems where SDL2 is installed in a way that expects `#include <SDL2/SDL.h>`.

Perhaps there is a way to patch this in a way that doesn't affect other builds.

qkdreyer:
It was required for the iOS hashlink build IIRC

tobil4sk:
Maybe we can make sure everything specific for the IOS builds is behind `#ifdef HL_IOS` blocks then. If you still have the dev environment, do you know where the include path is for SDL is when building for IOS? It is odd that it requires `#include <SDL.h>` instead of `#include <SDL2/SDL.h>`.

PR #361, opened 2020-04-11
Can you add build static library for code.c, jit.c, profiler.c and etc

malikzh:
Hi! I making binding for Go, for my purposes. Can you add target to build hl static library?

Go bindings for hashlink:https://github.com/malikzh/gohl


Makefile
```diff
@@ -206,10 +212,13 @@ release_osx:
 .c.o :
 	${CC} ${CFLAGS} -o $@ -c $<
 
+clean_a:
+	rm -f hls.a
```
Aurel300, line 216: Should be prefixed with `lib`.

Aurel300:
Why `libhls.a` and not `libhl.a`? You can still refer to the statically-built version by using the full path.

Makefile
malikzh: agree

malikzh:
> Why `libhls.a` and not `libhl.a`? You can still refer to the statically-built version by using the full path.

Fixed

Aurel300:
Why close?

PR #367, opened 2020-04-23
Add debug support for MacOS

rcstuber:
This PR adds Hashlink debugging support for OSX/MacOS. Things done:

- [x] Wrote a utility library for Mach Task debugging (`mdbg.c`) inspired by MacDBG and other sparsely scattered resources on the web (there's no documentation by Apple!)
- [x] Implemented required functions in `debug.c`
- [X] Implemented a few other things (e.g. `hl_debug_break()`)
- [x] ~~Adjusted Makefile to also build and install a native shared lib (`libhldebug.dylib`) which will be used by the Node version of the debugger.~~
- [x] Tested it as much as possible

**Known issues:**
- [ ] Currently it's not possible to edit break flags (e.g. Break on all exceptions), possibly due to some alignment issues or 4 bytes values or I don't know (@ncannasse kindly have a look)
- [ ] No 32bit support, but I believe all Macs are 64bit these days

Feel free to comment and **please help test it!** I will add a link to the VSCode/Node version PR below.

Debugger/VSCode Extension: https://github.com/vshaxe/hashlink-debugger/pull/78


Aurel300:
Why close?

ncannasse:
Thanks ! I will review asap, some question though : why having a separate libhldebug ? usually debug.c is linked directly into libhl.

rcstuber:
The libhldebug.dylib is for the Node/VSCode debugger version only, it’s basically debug.c as a standalone version (as we discussed). With the HL version it‘s statically linked in of course.
It made no sense to me to pull the source files over to the hashlink debugger project somehow and build them from there (since debug.c is part of Hashlink’s core functionality).

ncannasse:
Could you not load the primitives from libhl directly?

rcstuber:
:-| It never crossed my mind! Let me have a look... If it works, all the better, the less changes the better

rcstuber:
Ok, I didn't see forest for the trees, `libhl` is totally sufficient. I removed libhldebug completely! 

I also bumped the the minor version +1. If you want to just bump the patch version, let me know. In the VSCode extension I adjusted the guard to notify users of an incompatible hl version by checking if it's > "1.11.0", so I guess there won't be an in-between release and that's sufficient for now.

ncannasse:
Thanks !

PR #368, opened 2020-04-26
build: add codesign_osx make target

qkdreyer:


ncannasse:
Can you move entitlement.xml file to other directory?

rcstuber:
entitlements.xml could be moved to `/include/mdbg`, unless there should be a new conf folder of sorts.

Also maybe the expiration period should also be higher than 365 days?

qkdreyer:
I've moved it in `other/osx` folder, `include/mdbg` felt wrong since this file shouldn't be included while compiling

qkdreyer:
I don't mind increasing the period, but what value should we choose ?

ncannasse:
Oops maybe the period should have been extended before merging :-/

qkdreyer:
I think it's some false positive errors, because sudo isn't enough, you have to enter your credentials in order to be able to codesign

rcstuber:
Nicolas is on a merging streak 😅 I haven’t been able to test it yet, but looks like now it‘s re-creating the self-signing certificate every time we sign the app, instead of creating it when needed. But I guess, if it overwrites the certificate that‘s fine, as long as it doesn‘t add new keys and clutters the keychain – will check once at my laptop.

PR #373, opened 2020-04-27
minor improvements to OS X installation routine

rcstuber:
This tiny PR fixes two issues related to codesigning in OSX:

1) The "hl-cert" signing identity is now removed prior to generating a new certificate, rather then after. This prevents duplicate certificates of the same name – which is actually possible – being created when the command fails before the signing identity could be removed in the old version.

2) There is an issue with MacOS's security during `make install` which does not allow replacement of differently signed `hl` executables in `/usr/local...`. We therefore need to explicitly uninstall and then install to make this possible and prevent signing corruption

@qkdreyer @ncannasse


ncannasse:
There's still some problem with code signing in CI. Can you check ?

qkdreyer:
I found the issue, will push the fix asap

qkdreyer:
https://github.com/rcstuber/hashlink/pull/1

rcstuber:
What was the issue exactly? The permissions popup window or the azure pipeline failing or something we didn't notice yet?

qkdreyer:
@rcstuber There were 2 problems :
- The codesign utility was opening a GUI promp asking for user and password in order to allow the private key to be used
- I assued that we could test the codesign using cmake, but that last one rewrite the Makefile and removes this target

PR #377, opened 2020-04-29
use $(LIBEXT) instead of manually setting .hdll everywhere

jdonaldson:
I'm working through creating a hashlink formula for Homebrew: https://github.com/Homebrew/homebrew-core/pull/53790  

One of the CI checks is to validate that any generated libs have known MacOs-related extensions (.dylib, .framework, etc).  The current installation script for hashlink uses a platform specific "LIBEXT" variable for the main libhl library, but the supporting libraries  (openal, ssl) all use .hdll (a default for windows).  

This PR uses the LIBEXT extension for *all* installed libraries.  I tested this on MacOs, and verified the correct behavior for HomeBrew, but I haven't tested it anywhere else.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,11 +117,11 @@ install:
 	mkdir -p $(INSTALL_DIR)/include
 	cp hl $(INSTALL_DIR)/bin
 	cp libhl.${LIBEXT} $(INSTALL_DIR)/lib
-	cp *.hdll $(INSTALL_DIR)/lib
+	cp *.$(LIBEXT) $(INSTALL_DIR)/lib
 	cp src/hl.h src/hlc.h src/hlc_main.c $(INSTALL_DIR)/include
 
 uninstall:
-	rm -f $(INSTALL_DIR)/bin/hl $(INSTALL_DIR)/lib/libhl.${LIBEXT} $(INSTALL_DIR)/lib/*.hdll
+	rm -f $(INSTALL_DIR)/bin/hl $(INSTALL_DIR)/lib/libhl.${LIBEXT} $(INSTALL_DIR)/lib/*.$(LIBEXT)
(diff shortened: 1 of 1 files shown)


Aurel300:
This is a bit strange. `hdll` is a HashLink-specific extension that is used on any OS. It is not Windows-specific (that would just be `dll`), and is similar to Neko's `ndll`. Also, have you actually tested this working when loading anything from libraries? I don't think it can, since you have not changed this:

https://github.com/HaxeFoundation/hashlink/blob/a30209ef5020fce4b3b1cda3e7a4fed2edd7f118/src/module.c#L351

nanjizal:
I do wonder if rather than Haxe use brew for building packages and risk that older osx are not supported, and brew problems and  update aspects, it would be ideal to have some ninja, especially for setting up ocaml and building haxe but also for hashlink.
https://ninja-build.org/   textmate2 uses ninja and it is one of the easiest complex projects to setup and get working on a mac.
Anyway this is not a request for official ninja over brew support just thought it might be something that may interest you, and may work out simpler?

jdonaldson:
> Also, have you actually tested this working when loading anything from libraries? I don't think it can [...]

Thanks, I missed this.

jdonaldson:
I think it's best if we get some sort of exception or provide some sort of whitelist for these hashlink libraries.  I'll leave this PR up for more discussion.

ncannasse:
Is it just a warning ? If it doesn't prevent the brew to be build I think we should just skip it.

jdonaldson:
I'm going to cancel this, I don't think modifying the hashlink library naming pattern is a good idea for now.

jdonaldson:
> Is it just a warning ? If it doesn't prevent the brew to be build I think we should just skip it.

Yeah, it's a warning.  Still, homebrew is a huge project, and I've put effort into the hashlink formula so as to reduce noise as much as possible.

PR #381, opened 2020-05-16
Rudimentary clipboard support

muchitto:
Rudimentary clipboard support. ~~Only supports setting and getting text from the clipboard at the moment (as that is probably the most likely use case and that is the use case as I needed it for), but could support images and other formats in the future.~~ It supports, getting and setting text and also supports getting image data from and to the clipboard.

It uses the dacap/clip c++ library (https://github.com/dacap/clip/), due to its ease of use.

Don't know if this is the right way to implement this (I'm using CMake, so don't know about the Makefile), but here goes. It is now an additional library (hlclip). Is this something that should be a core Hashlink functionality?

Usage atm:
```haxe
os.Clipboard.setText("test");
os.Clipboard.getText();
```

(Related issue: https://github.com/HeapsIO/heaps/issues/520)


Brewfile
```diff
@@ -10,3 +10,4 @@ brew "openal-soft"
 brew "mbedtls"
 brew "libuv"
 brew "openssl"
+brew "unzip"
```
Aurel300, line 13: Is this really necessary? `unzip` is a built-in command on OS X.

libs/CMakeLists.txt
```diff
@@ -21,6 +21,11 @@ if(WITH_FMT)
     add_subdirectory(fmt)
 endif()
 
+option(WITH_CLIP  "Build clip.hdll" ON)
```
Aurel300, line 24: You call it `clipboard.hdll` later.

libs/clip/clip/Clipboard.hx
```diff
@@ -0,0 +1,20 @@
+package clip;
```
Aurel300, line 1: A whole top-level package just for clipboard? :/

libs/clip/clip/Clipboard.hx
```diff
@@ -0,0 +1,20 @@
...
+    }
+    
+    static function set_clipboard_text (text : hl.Bytes) : Bool {
+        return false;
+    }
+
+    public static function setText (text : String) {
+        return @:privateAccess set_clipboard_text(text.toUtf8());
+    }
+}
```
Aurel300, line 20: Formatting inconsistent with Haxe stdlib, also a `Bytes` version should be even simpler than text.

Aurel300:
This seems like a relatively large change (additional hdll, dependency, and top-level package) for a relatively small benefit. Re: the Heaps issue, why not solve it with SDL's existing [clipboard functions](https://wiki.libsdl.org/CategoryClipboard) (+ similar Windows-specific API for hldx)?

Brewfile
muchitto: Yeah, probably not. I'm using Windows and haven't tested it on a Macbook yet.

libs/clip/clip/Clipboard.hx
muchitto: Yeah, thought that seems like a bit much. I kinda just followed the sdl -libs scheme and added it's own. Also, in the future you probably want more clipboard functionality (images etc), so there probably won't be only text related functionality there...

What would you suggest for the package name?

libs/CMakeLists.txt
muchitto: Yeah, that's because of the way I'm including the library (with CMake's add_subdirectory atm, don't know if that's the best way, but there isn't binary distributions for the library, so I thought I'd just attach it to the CMake build process itself). It creates a conflict in the build, because you have a subproject with the same name as the hl lib -project. I don't know if you can bypass that somehow, though.

muchitto:
Yeah, I actually tried the SDL Clipboard first, and it's good enough if you only want text, but I thought there is a pretty big possibility, that in the future you also want to use it for other types than text as well, so I just brought in that library as it brings a no hassle cross platform wrapper for those things. 

I don't actually know in what package this kind of functionality should belong in, though. I agree that it's kind of too much to add a new hdll just for clipboard. Should there be a kind of "OS" -library or something built in to Hashlink for these kinds of trivial little functionalities?

libs/CMakeLists.txt
muchitto: Ah, yeah, had a brainfart, finally looked at what you were actually pointing at :) Yeah that was from the time I called it "clip.hdll" and not "clipboard.hdll".

libs/os/os.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+HL_PRIM bool HL_NAME(set_clipboard_text)(const char * text) {
+    return clip::set_text(text);    
+}
+
+HL_PRIM vbyte * HL_NAME(get_clipboard_image_data)() {
+    clip::image img;
+    if(clip::get_image(img)) {
+        int size = img.spec().width * img.spec().height * (img.spec().bits_per_pixel / 8) * sizeof(char);
+        vbyte * data = (vbyte *)hl_gc_alloc_raw(size);
```
muchitto, line 20: I don't know, do you have to use hl_gc_alloc... functions, so that the memory is freed at some point or if we are returning bytes from the function, will it take care of them automatically?

Aurel300: This usage is correct. The `vbyte*` (= `hl.Bytes`) block is known to the GC and will be collected when no longer reachable.

libs/clip/clip/Clipboard.hx
muchitto: Should the haxe-formatter be used on the haxe files or what is the usual way of making sure, the formatting is correct?

libs/os/os.cpp
muchitto: Ok, thanks. :)

CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ set(HL_VERSION ${HL_VERSION_MAJOR}.${HL_VERSION_MINOR}.${HL_VERSION_PATCH})
 
 cmake_policy(SET CMP0042 NEW)
 
-if(WIN32)
+if(WIN32 OR WITH_OS)
     project(hashlink C CXX) # C++ required for directx
```
Aurel300, line 11:
```suggestion
    project(hashlink C CXX) # C++ required for directx and clipboard support
```

muchitto:
I haven't messed around with makefiles that much, I guess in it's current state the Hashlink makefile expects the libraries to be precompiled already, so I don't know about the clip library then? Should the library also be expected to be precompiled and put into specific folders under the `include/clip` or whats the deal?

Out of curiosity, why is there a makefile AND a cmake file?

Aurel300:
No, the makefile can be used to build the project without using cmake at all. I personally build HL like that on OS X. So please check how the makefile build process works for any of the other libraries, and make sure that the clipboard library can be built with a purely makefile-based workflow.

muchitto:
Sure, that can be done. 

It just wasn't that clear why there are two separate ways of building the project (which also have to be maintained separately of course). Oh and also a solution file for VS. So I guess that makes it three. :)

ncannasse:
Sorry for being late to the party.

I think we should directly use platform native code for clipboard handling, looking at daclip sources if necessary for tips. Also, this should be part of the `ui` library and not a new library.

PR #382, opened 2020-05-18
[UNIX] Don't crash on SIGPIPE

kLabz:
Currently, when running on linux (didn't test on macos) if remote socket hangs up the whole application silently crashes with SIGPIPE (and no stacktrace). I traced it all the way to libuv that does this on purpose for performance reasons (see https://github.com/joyent/libuv/issues/1254#issuecomment-41185781). The recommended (as I understood it) way to prevent crashes is to redirect SIGPIPE to SIG_IGN and handle socket errors in user code instead.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -125,6 +125,7 @@ static void setup_handler() {
 	act.sa_handler = handle_signal;
 	act.sa_flags = 0;
 	sigemptyset(&act.sa_mask);
+	signal(SIGPIPE, SIG_IGN);
 	sigaction(SIGSEGV,&act,NULL);
 	sigaction(SIGTERM,&act,NULL);
 }


ncannasse:
I'm not fan with the fix. Could we instead use SO_NOSIGPIPE on socket? It's more specific.

kLabz:
From what I read on all sources I can find, SO_NOSIGPIPE doesn't exist on linux, and SIGPIPE needs to be ignored manually.

Other sources:
https://stackoverflow.com/questions/26752649/so-nosigpipe-was-not-decleared#26752697
https://stackoverflow.com/questions/44551990/libuv-how-to-ignore-sigpipe-when-connection-reset-by-peer-linux

PR #384, opened 2020-05-24
[SDL] Add clipboard text API

kLabz:
Expose `SDL_HasClipboardText`, `SDL_GetClipboardText` and `SDL_SetClipboardText` in `sdl.hdll`

### DIFF
--- a/libs/sdl/sdl.c
+++ b/libs/sdl/sdl.c
@@ -781,3 +781,21 @@ DEFINE_PRIM(_CURSOR, cursor_create_system, _I32);
 DEFINE_PRIM(_VOID, free_cursor, _CURSOR);
 DEFINE_PRIM(_VOID, set_cursor, _CURSOR);
 DEFINE_PRIM(_ARR, get_devices, _NO_ARG);
+
+// clipboard
+
+HL_PRIM bool HL_NAME(has_clipboard_text)() {
+	return SDL_HasClipboardText();
+}
+
+HL_PRIM const char *HL_NAME(get_clipboard_text)() {
+	return SDL_GetClipboardText();
+}
+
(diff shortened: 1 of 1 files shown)


libs/sdl/sdl.c
```diff
@@ -781,3 +781,21 @@ DEFINE_PRIM(_CURSOR, cursor_create_system, _I32);
...
 DEFINE_PRIM(_ARR, get_devices, _NO_ARG);
+
+// clipboard
+
+HL_PRIM bool HL_NAME(has_clipboard_text)() {
+	return SDL_HasClipboardText();
+}
+
+HL_PRIM const char *HL_NAME(get_clipboard_text)() {
+	return SDL_GetClipboardText();
```
Aurel300, line 792: Looks like a memory leak to me, the docs say this pointer must be freed with `SDL_free`. We want to copy the bytes into GC-managed memory, then free the original pointer immediately.

kLabz: Not sure how tbh, I'm not used to managed memory.

Aurel300: Something like this might work?

```suggestion
	char *text = SDL_GetClipboardText();
	if (!text)
		return NULL;
	vbyte *ret = (vbyte *)hl_to_utf16(text);
	SDL_free(text);
	return ret;
```

Then the returned value should be directly usable for a string, since it is UTF-16 already.

ncannasse:
Seems weird that you convert to utf16 in get but not from utf16 in set ;)

kLabz:
I wasn't doing it initially. And now it crashes when I try to do what you ask.
Nevermind then.

Aurel300:
@kLabz please share the code you tried, it may have been a simple mistake ^^ There is no reason that a UTF-16 to UTF-8 conversion should crash if done correctly.

PR #387, opened 2020-06-02
Allow passing extra linker flags

vladimyr:
This is required by Homebrew https://github.com/Homebrew/homebrew-core/pull/53790#issuecomment-637662089 because it installs hashlink's `hddl`-s outside of `/usr/local/lib`.


vladimyr:
Merci ♥️

PR #401, opened 2020-07-15
feat: add window winError

qkdreyer:


ncannasse:
In which case is this needed ?

PR #424, opened 2020-11-26
Add null option for library in MakeFile

PXshadow:
Quickly be able to set LIBS=null to have no libraries be generated.


PXshadow:
closing as fmt is almost always needed from what I understand.

PR #432, opened 2021-01-04
add broadcast ability to udp sockets

rutenkolk:
Hello,

Trying out heaps.io, I noticed, that the implementation for sockets was missing the `setBroadcast` function for udp.

I hope I implemented this correctly as it isn't a big diff. 

I tested sending broadcasts with ubuntu 20.04, haxe 4.1.4, heaps 1.8.0
They showed up in wireshark and I could recieve them on another machine in the same network.


RutenkolkC:
If I see this correctly, the build failure on windows is unrelated to this PR. Seems like openal cannot be installed on windows.

I don't understand what is going on, but I assume the CI would need to be force-rerun at some point (assuming it's the fault of openal)

PR #435, opened 2021-01-07
Added method winGetPixelSize to SDL, for getting pixel size of high DPI windows

jefvel:
Related to https://github.com/HeapsIO/heaps/issues/589


ncannasse:
Why calling it win_get_pixel_size while its drawable_size in SDL ?

jefvel:
that makes more sense. I'll rename it

PR #444, opened 2021-02-20
SDL/DX: Add mouse handling API

Yanrishatum:
* DirectX: Added `Event.mouseXRel` and `mouseYRel` variables, fixing the data mismatch between DX and SDL.
* DirectX: Added `setCursorPosition` and `setCursorPositionGlobal` to set mouse position with relative to window and screen coordinates.
* DirectX: Added support for relative mouse mode via `setRelativeMouseMode` and `getRelativeMouseMode` (see below)
* DirectX: Fixed `winClipCursor` and improved how cursor capture works.
* SDL: Added `Window.warpMouse` and `Sdl.warpMouseGlobal` bindings to set the mouse position.
* SDL: Added `Sdl.getRelativeMouseMode`.
* SDL: Added `Window.grab` that traps the cursor within the window boundaries.

In short, for SDL: Added bindings for common mouse handlers such as setting position, relative mode and cursor trap.

DirectX implementation is heavily based on how SDL does it, especially improvement of cursor clipping, so it would be possible to drag the window and click the title buttons while the cursor grab is enabled by unfocusing the window and licking on the titlebar. 
Relative mouse mode enables raw input mode, and also implies the trapping of cursor within the window. When enabled, mouse is hidden, but still moves within the window boundaries.

This is a prerequisite PR to implement https://github.com/HeapsIO/heaps/issues/568


Yanrishatum:
Fixed merge conflicts and re-tested (it worked back then - it works now). The Heaps PR will be redone _after_ this one is merged due to mouse lock being added to JS and I'm not dealing with figuring out how to combine it just to see it rot for another 2 years. @ncannasse
It sat here for more than 1.5 years without as much as a comment, I have full rights to be salty. :^) 🧂

PR #453, opened 2021-03-19
fix cleanup for main

### LINKED ISSUE #452: Assert error from allocator after multiple reinitializations
This is about [godot-haxe](https://github.com/vonagam/godot-haxe). Godot reloads gnative libraries every time editor gets focus. Reload means that you should first free everything and then initialize again (this way updates to library registered classes are applied immediately).

On third initialization of the library an error from [this line](https://github.com/HaxeFoundation/hashlink/blob/c2786410a0fc525edda88b266fee1158efc38f44/src/allocator.c#L203) in `allocator.c` is thrown. 

I don't understand much what is going on in allocator or how to properly debug it. Is there any clue by any chance that can narrow down a reason for the failure?

(Functions for `init` and `free` of hashlink's jit are [here](https://github.com/vonagam/godot-haxe/blob/development/sources/c/embed_jit.c). This is basically `main` from `main.c`.)

vonagam:
Closes #452 (at least i cannot reproduce the failure with 3 reloads anymore).

Two things so far that i found that make cleanup in main not complete:

- In `main` there is a call to [`hl_register_thread`](https://github.com/HaxeFoundation/hashlink/blob/c2786410a0fc525edda88b266fee1158efc38f44/src/main.c#L207), but no call to `hl_unregister_thread` at the end. (For `main` itself it is ok since it does not care much about full cleanup before exit, but it can be used as a reference point for proper integration and better to make it right.)

- In `hl_module_init` globals are passed to [`hl_add_root`](https://github.com/HaxeFoundation/hashlink/blob/c2786410a0fc525edda88b266fee1158efc38f44/src/module.c#L521), but in `hl_module_free` globals [are freed](https://github.com/HaxeFoundation/hashlink/blob/c2786410a0fc525edda88b266fee1158efc38f44/src/module.c#L763) without being removed as gc roots.

### DIFF
--- a/src/gc.c
+++ b/src/gc.c
@@ -762,6 +762,12 @@ static void hl_gc_init() {
 #	endif
 }
 
+static void hl_gc_free() {
+#	ifdef HL_THREADS
+	hl_remove_root(&gc_threads.global_lock);
+#	endif
+}
+
 // ---- UTILITIES ----------------------
 
 HL_API bool hl_is_blocking() {
@@ -806,6 +812,7 @@ void hl_global_init() {
 
 void hl_global_free() {
 	hl_cache_free();
+	hl_gc_free();
 }
 
 struct hl_alloc_block {
--- a/src/main.c
+++ b/src/main.c
@@ -244,6 +244,7 @@ int main(int argc, pchar *argv[]) {
 	}
 	hl_module_free(ctx.m);
 	hl_free(&ctx.code->alloc);
+	hl_unregister_thread();
 	hl_global_free();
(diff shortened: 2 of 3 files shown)


vonagam:
Another (seems like last one) unremoved root is from [`hl_gc_init`](https://github.com/HaxeFoundation/hashlink/blob/c2786410a0fc525edda88b266fee1158efc38f44/src/gc.c#L761) for global lock. It is allocated and added in `hl_global_init`, but not removed in `hl_global_free`.

PR #458, opened 2021-06-16
Added SDL and win32 bindings to get/set system clipboard

Azrou:


ncannasse:
I think we need also the primitives in `ui_stub.c` for not windows not gl platforms (consoles)

runeknytling:
It looks like DirectX has a similar issue

Issue #73, opened 2017-09-18
Nighlty builds


RealyUniqueName:
Is it possible to setup nightly builds just like compiler nightlies?
I think that will allow people to avoid all issues they have on Linux and OSX trying to build hashlink manually.

ncannasse:
We need to fix our CI first (ping @andyli) then see if @waneck can help with it. Given that we do that already for Neko it shouldn't be that hard.

andyli:
CI is green at the moment.

waneck:
Just let me know what files we should submit and I'll set it up

ncannasse:
Ah good :) I don't know where the build files are located but for a full distrib we need:

- hl binary (.exe in windows)
- libhl (.so or .dll depending on target)
- *.hdll files
- we might want to include some binaries on Windows so the package does not need any extra install (SDL2.dll, OpenAL32.dll -- openal soft version, vs_redist.x86.exe)
- sources necessary to build hl/C : hlc_main.c , hl.h, hlc.h  (+.lib files on windows)

You can check my windows distribution here for an example 
https://github.com/HaxeFoundation/hashlink/releases

waneck:
Do we have hashlink builds setup on AppVeyor for windows? It would be nice if we had a `make release` that would at least copy all the relevant files. But I'll take a look into this tonight

RealyUniqueName:
What about hlsdl and friends? Should they also be in nightlies?

ncannasse:
@waneck Appveyor is green, can we setup the nightly ? :)

ncannasse:
@waneck up :)

nadako:
Note that I did set up build artifacts on appveyor: https://ci.appveyor.com/project/HaxeFoundation/hashlink/build/498/job/ih1ilw6fkfaagmjq/artifacts

ncannasse:
Anyone up to setup nightly builds ? :) @waneck @andyli

ncannasse:
Ping again @andyli  :)

thomasjwebb:
This would be useful to have. I'm trying to add hashlink to a docker config so it would be better to have it download and install something than try to build it.

thomasjwebb:
I ended up just having docker compile and install but in general it would be useful to have builds and ultimately packages.

Simn:
I'm told that this can be closed now!

PR #462, opened 2021-08-01
Add GitHub Actions workflow

sebthom:
This workflow runs on each commit/PR to verify it. When a commit is made to the main branch a GitHub pre-release named "latest" is created/updated containing binaries for linux/mac/win based on the last commit.

This also addresses #73

You can find an example here: https://github.com/sebthom/hashlink/releases/tag/latest

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+      day: monday
+      time: "10:00"
+    commit-message:
+      prefix: fix
+      prefix-development: chore
+      include: scope
+    labels:
+      - dependencies
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,345 @@
+# https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions
+name: Build
+
+on:
+  push:
+  pull_request:
+  workflow_dispatch:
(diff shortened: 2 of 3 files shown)


Apprentice-Alchemist:
I would suggest using make for linux/osx, as cmake doesn't generate `ssl.hdll` currently.
Another issue is that cmake does not codesign the binaries on osx, meaning you wouldn't be able to use the debugger with them.
(Here's how I do it : <https://github.com/Apprentice-Alchemist/hashlink/blob/feature/github-ci/.github/workflows/main.yml>.)

sebthom:
@Apprentice-Alchemist I don't mind to use make or cmake. what I like about the cmake approach is that it produces the final zip/tar.gz archives and I don't have to repeat the logic of how to package hashlink again.

Maybe the cmake config can be fixed somehow to sign macos binaries and generate the `ssl.hddl`?

Apprentice-Alchemist:
On linux/macos you can also do `make release` to package binaries generated with `make`.

ncannasse:
Hi ! Thanks for the contribution !
Yes I mostly maintain the `make` build, cmake was added as an alternative but is not entirely supported, so if you can switch to make I would be happy to merge it.

sebthom:
I changed to make for linux/darwin and msbuild on windows.

An example of the generated artifacts can be found here https://github.com/sebthom/hashlink/releases/tag/latest

@Apprentice-Alchemist thanks for your build example, very helpful!

sebthom:
@ncannasse could you please review and possibly merge this PR?

sebthom:
@ncannasse could you please review and possibly merge this PR?

ncannasse:
Merged ! thanks for the contrib !

sebthom:
@ncannasse thanks for merging! apparently some things on the github runners as well as in the project config has changed since I created this GHA build config. I am going to provide another PR adjusting it accordingly

PR #465, opened 2021-08-09
I64 implementation

RealyUniqueName:
Required for HaxeFoundation/haxe#10347
I implemented it as a separate set of functions to avoid intrusion in hashlink internals, which I don't understand yet.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,6 +79,7 @@ file(GLOB std_srcs
     src/std/ucs2.c
     src/std/thread.c
     src/std/process.c
+    src/std/num.c
 )
 
 if(CMAKE_SYSTEM_NAME MATCHES "Darwin")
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -157,7 +158,7 @@ endif()
 if(BUILD_TESTING)
 
     find_program(
-        HAXE_COMPILER
+        HAXE_COMPILERs
```
Aurel300, line 161: This seems to be a typo?

RealyUniqueName: Thanks

ncannasse:
I think it's better to use directly JIT operations for that. HL already has a native I64 type, exposed through `hl.I64`
However I haven't implemented the JIT operations on 64 bits integers. The main reason was that support for 32 bits was quite difficult to implement. 

I think this could work if could first focus on adding 64 bit JIT support (through `hashlink/src/jit.c`) , then we can find a way to use such primitives fallback at JIT level for 32 bit support.

ncannasse:
Int64 is fully supported now, closing.

Issue #271, opened 2019-07-22
Mingw64 support ?


melMass:
Hi,

Whenever I try to build HL from source on windows Mingw64 I get the following error:

```ruby
hashlink\src\std\buffer.c:377:3: error: unsupported non-standard concatenation of string literals
  377 |   hl_buffer_str_sub(b, buf, usprintf(buf, 32, _PTR_FMT USTR("H"),(int_val)v));
      |   ^~~~~~~~~~~~~~~~~
mingw32-make[2]: *** [CMakeFiles\libhl.dir\build.make:273: CMakeFiles/libhl.dir/src/std/buffer.c.obj] Error 1
mingw32-make[1]: *** [CMakeFiles\Makefile2:310: CMakeFiles/libhl.dir/all] Error 2
```

ncannasse:
If you simply remove the USTR("H") do you get some other errors?

melMass:
Sorry I was busy. I'm now back, i'll try your suggestion and report back later today.

melMass:
Removing USTR did solved the original issue. Now it complains about missing library references:
```ruby
src/std/socket.o:socket.c:(.text+0x128): undefined reference to `__imp_socket'
src/std/socket.o:socket.c:(.text+0x161): undefined reference to `__imp_socket'
src/std/socket.o:socket.c:(.text+0x199): undefined reference to `__imp_inet_addr'
src/std/socket.o:socket.c:(.text+0x1ba): undefined reference to `__imp_gethostbyname'
src/std/socket.o:socket.c:(.text+0x1e3): undefined reference to `__imp_inet_ntoa'
src/std/socket.o:socket.c:(.text+0x216): undefined reference to `__imp_gethostbyaddr'
src/std/socket.o:socket.c:(.text+0x258): undefined reference to `__imp_gethostname'
src/std/socket.o:socket.c:(.text+0x2ee): undefined reference to `__imp_accept'
src/std/socket.o:socket.c:(.text+0x358): undefined reference to `__imp_getpeername'
src/std/socket.o:socket.c:(.text+0x36e): undefined reference to `__imp_ntohs'
src/std/socket.o:socket.c:(.text+0x3b8): undefined reference to `__imp_getsockname'
src/std/socket.o:socket.c:(.text+0x3ce): undefined reference to `__imp_ntohs'
src/std/socket.o:socket.c:(.text+0x427): undefined reference to `__imp_setsockopt'
[...]

PR #482, opened 2021-10-23
Fix MinGW build

Koncord:
Should fix Issue #271.


N3rdL0rd:
I'm having some issues compiling std/thread.c on MinGW even with these changes, since it uses __try {...}.

tobil4sk:
@N3rdL0rd Here is an updated PR to fix mingw support, it should solve this issue: #831.

PR #483, opened 2021-10-29
Fix linux dynamic library issue when installing to `/usr/local/`

### LINKED ISSUE #386: Linux Mint "make install" issue with libhl.so being installed into "/usr/local/" and not being seen by HashLink
> hl: error while loading shared libraries: libhl.so: cannot open shared object file: No such file or directory

On Linux Mint 19.3 Cinnamon 64bit, HashLink 1.11
Issue: `make install` installs `libhl.so` into `/usr/local/lib` and HashLink doesn't see it.
Workaround: use `make INSTALL_DIR=/usr install` instead as suggested in [#221 ](https://github.com/HaxeFoundation/hashlink/issues/221#issuecomment-455892192)

tobil4sk:
Resolves #386

The way this is solved is by modifying rpath settings. The hl executable will search its original path first (for relocatable builds, for testing and development etc) then the installation library path. Addresses both make and cmake.

It might be useful to have a setting to switch between the relocatable build and the installable build, however for now this works fine.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -145,6 +145,10 @@ add_executable(hl
 	src/profile.c
 )
 
+if (UNIX AND NOT APPLE)
+    set_target_properties(hl PROPERTIES INSTALL_RPATH "$ORIGIN;${CMAKE_INSTALL_PREFIX}/lib")
+endif()
+
 target_link_libraries(hl libhl)
 
 if(WIN32)
--- a/Makefile
+++ b/Makefile
@@ -89,7 +89,7 @@ else
 
 # Linux
 CFLAGS += -m$(MARCH) -fPIC -pthread -fno-omit-frame-pointer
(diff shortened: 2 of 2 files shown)


tobil4sk:
This may be relevant:
https://github.com/HaxeFoundation/hashlink/pull/437#issuecomment-757533270
> For Makefile: it's actually intended the way it is atm, I don't think you should add paths outside of the current directory for security reasons, as it might load untrusted binaries with privileges without notice.

tobil4sk:
@ncannasse Is there any chance of getting this merged? This problem completely breaks hashlink for anyone using it on Linux, unless they install it in `/usr/`, which is not where it goes by default and is meant to be reserved for package managers, or unless they modify their system to work around the issue, which really shouldn't be necessary just to use one piece of software. New users frequently come on the Discord asking about how to fix this.

ncannasse:
I'm not sure if that's the best solution but well people can always change the Makefile so better having it work by default.

PR #490, opened 2021-11-11
Implement profiler for linux.

### LINKED ISSUE #402: Implement profiler output for Linux
Feature request for the profiler to be included in the Linux version, as mentioned by the wiki it's windows only: 
`Profiler is introduced in HL 1.11 and only implemented for Windows so far, see src/profile.c to contribute other platforms.`
So the only thing left for the profiler to work on Linux is translating the `#ifdef HL_WIN` parts of `profile.c` or is there something else?

Apprentice-Alchemist:
Closes #402.

### DIFF
--- a/src/profile.c
+++ b/src/profile.c
@@ -22,6 +22,12 @@
 #include <hl.h>
 #include <hlmodule.h>
 
+#ifdef HL_LINUX
+#include <signal.h>
+#include <semaphore.h>
+#include <unistd.h>
+#endif
+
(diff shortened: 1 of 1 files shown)


ncannasse:
Thanks !

ncannasse:
Were you able to make it work entirely ?

Apprentice-Alchemist:
It produced correct results when I tested it with the heaps samples, so it seems to work entirely.

PR #505, opened 2021-12-22
Fix Visual Studio project templates

tobil4sk:
The templates generated by the `hashlink` haxelib broke after #486 so this is required to get them working again.

I added a vs2019 build target to the hl solution as it's needed for the vs2019 template to work. Then I updated the template files to look in the correct directories.

I also added the vs2017 template as it was missing.


ncannasse:
Hello,
when compiling with VS2019 we noticed a very significant slowdown of the VM. We still haven't had time to investigate the cause of this so I prefer we don't add anything related to vs 2019 atm.

tobil4sk:
Alright, understood. I will remove the vs2019 parts of this pull request and just focus on fixing the vs2015 template and adding the vs2017 template.

I'll also remove the vs2019 template and create a new PR for adding the vs2019 build target as well as the template for when those issues are addressed.

ncannasse:
There seems to be a line ending difference in your commit since all lines of some templates are marked as modified, could you fix?

PR #508, opened 2022-01-06
Wrapped filesystem events support from libuv

Azrou:
The hx side implementation is in a HaxeFoundation/haxe PR


ncannasse:
I think we should pass the callback at init_wrap since we have a single callback for all paths.
Otherwise there's a thread question, as we are not allowed to perform GC allocations in a thread that has not been properly registered.

PR #512, opened 2022-01-31
SDL: Add drag&drop support

Yanrishatum:
* Add handling of drag&drop events from SDL
* Add flag to enable and disable drag&drop.

It seems SDL have drag&drop enabled by default, and master branch have a memory leak every time someone drops a file onto the window because SDL also expects user to manually free the dropped file contents.

Related: heapsio/heaps#1043


Yanrishatum:
@ncannasse 
We'd really like to get that merged. This PR is primarily for development, because we do not use HIDE (for many, MANY reasons) and make our own editing tools using Imgui. And being unable to drag&drop files is a massive inconvenience for the development (along with #559 ). 

This touched up explicitly only for SDL, because getting Drag&Drop working on DX is much more involved, and we're fine with using only SDL for development. I will revise the Heaps integration of Drag&Drop PR after this PR is merged so I can properly test it.

ncannasse:
Could you add @:native("?sdl","xxxxx") in front of the new functions on haxe side so that someone that compiles with the new header but with the old binaries will be able to run their program as long as they don't use them ? once done I can merge, no pb.

Yanrishatum:
Done.

ncannasse:
Thanks

Issue #525, opened 2022-03-03
Linux nightly build is unusable


tobil4sk:
Right now, the nightly build for linux cannot be used "out of the box" because it is not relocatable. With neko you were able to simply download a nightly build, extract it, add the extracted path to PATH, and use the neko binary from anywhere without any issues. When trying to do this with hashlink, `libhl.so` is not found, because the rpath used when building it does not contain `$ORIGIN`.

Ways to fix this:
- Adding `$ORIGIN` to rpath, but Nicolas has been reluctant to add anything to rpath, see: https://github.com/HaxeFoundation/hashlink/pull/437#issuecomment-757533270 
- Distribute the nightly build with an `install.sh` script. However, unless this script installs the binaries directly to `/usr/`, then the user will experience #386. This cannot be fixed either unless we modify the rpath setting.

Neko also had a `RELOCATABLE` cmake option, to toggle whether `$ORIGIN` is added to rpath.

Unless rpath is modified, then there is no way for Linux users to be able to use hashlink without installing to `/usr/`. This is especially confusing for new linux users or people just trying to find a way to test if their game works on linux with the latest hashlink.

This is a pretty urgent issue. I am happy to create a new PR for a fix for this, or modify #483 to fit the solution decided upon, but a decision has to be made on how to approach this.

Magnesus:
Not sure if related but I also get "./hl: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.30' not found (required by ./hl)
./hl: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by ./libhl.so)" errors when trying to run that hl on Ubuntu.

Apprentice-Alchemist:
Are you using Ubuntu 18.04?
If yes, then it's just the usual glibc issues.
CI uses 20.04, and thus a more recent version of glibc.
Solutions: build from source, or upgrade to an up to date OS.

Magnesus:
Yes, I use old Ubuntu. I have to admit though I was expecting HL to work on every Linux version out there, otherwise releasing with it is a bit risky since users may have unsupported version of libraries.

Apprentice-Alchemist:
> work on every Linux version

That is not practically feasible since it would require building against the oldest versions of glibc etc.

But, building on Ubuntu 18.04 should do the trick for people that for some reason use LTS releases of Debian or Ubuntu.

[...]

PR #529, opened 2022-03-20
Build with Ubuntu 18 and Windows 2019

sebthom:
This PR downgrades the GHA build from Ubuntu 20 to 18 to address #525. It also changes the windows runner from windows-2016 to windows-2019 as the former has been decomissioned.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,15 +22,13 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          # cannot build 32 bit binaries anymore:
-          # a) ubuntu-18.04 only supprots GLIBC 2.27 but 2.29 is required: https://www.digitalocean.com/community/questions/how-install-glibc-2-29-or-higher-in-ubuntu-18-04
-          # b) ubuntu-20.04 does no longer provide i386 packages
+          # TODO cannot build 32 bit binaries. compile errors
           #- target: linux-i386
           #  runner: ubuntu-18.04
           #  haxe_latest_dist: linux64/haxe_latest.tar.gz
 
           - target: linux-amd64
-            runner: ubuntu-20.04
+            runner: ubuntu-18.04
             haxe_latest_dist: linux64/haxe_latest.tar.gz
 
           - target: darwin
@@ -38,8 +36,8 @@ jobs:
             haxe_latest_dist: mac/haxe_latest.tar.gz
 
           - target: win32
-            runner: windows-2016  # has VS2017 preinstalled which supports PlatformToolset <= v141, WindowsTargetPlatformVersion 8.1
-            #runner: windows-2019 # has VS2019 preinstalled which supports PlatformToolset <= v142, WindowsTargetPlatformVersion 10
+            #runner: windows-2016 # has VS2017 preinstalled which supports PlatformToolset <= v141, WindowsTargetPlatformVersion 8.1
+            runner: windows-2019 # has VS2019 preinstalled which supports PlatformToolset <= v142, WindowsTargetPlatformVersion 10
             haxe_latest_dist: windows64/haxe_latest.zip
             ffmpeg_url: https://github.com/HaxeFoundation/hashlink/files/5648055/ffmpeg-3.4.2-win32-dev.zip
             msbuild_Configuration: ReleaseVS2017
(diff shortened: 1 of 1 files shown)


sebthom:
This should solve #525

sebthom:
@ncannasse I also tried to enable the 32bit builds on Ubuntu but compilation fails with 
```c
cc -Wall -O3 -I src -msse2 -mfpmath=sse -std=c11 -I include -I include/pcre -I include/mikktspace -I include/minimp3 -D LIBHL_EXPORTS -m32 -fPIC -pthread -fno-omit-frame-pointer -I /usr/include/i386-linux-gnu -o src/std/debug.o -c src/std/debug.c
src/std/debug.c: In function ‘get_reg’:
src/std/debug.c:202:43: error: ‘struct user_fpregs_struct’ has no member named ‘xmm_space’; did you mean ‘st_space’?
   case 11: return (void*)(-((int_val)&fp->xmm_space[0])-1);
                                           ^~~~~~~~~
                                           st_space
src/std/debug.c: In function ‘hl_debug_read_register’:
src/std/debug.c:349:22: warning: pointer targets in passing argument 2 of ‘hl_debug_read’ differ in signedness [-Wpointer-sign]
   hl_debug_read(pid, addr + (-((int_val)r)-1), (vbyte*)&out, sizeof(void*));
                      ^~~~
src/std/debug.c:104:13: note: expected ‘vbyte * {aka unsigned char *}’ but argument is of type ‘char *’
 HL_API bool hl_debug_read( int pid, vbyte *addr, vbyte *buffer, int size ) {
             ^~~~~~~~~~~~~
make: *** [src/std/debug.o] Error 1
Makefile:224: recipe for target 'src/std/debug.o' failed
```

Are 32 bit builds not supported anymore or am I missing something in the configuration?

Apprentice-Alchemist:
> This should solve #525

Partially. There's the still the rpath issue that needs to be addressed.

> I also tried to enable the 32bit builds on Ubuntu but compilation fails with
> 
> ```c
> cc -Wall -O3 -I src -msse2 -mfpmath=sse -std=c11 -I include -I include/pcre -I include/mikktspace -I include/minimp3 -D LIBHL_EXPORTS -m32 -fPIC -pthread -fno-omit-frame-pointer -I /usr/include/i386-linux-gnu -o src/std/debug.o -c src/std/debug.c
> src/std/debug.c: In function ‘get_reg’:
> src/std/debug.c:202:43: error: ‘struct user_fpregs_struct’ has no member named ‘xmm_space’; did you mean ‘st_space’?
>    case 11: return (void*)(-((int_val)&fp->xmm_space[0])-1);
>                                            ^~~~~~~~~
>                                            st_space
> src/std/debug.c: In function ‘hl_debug_read_register’:
> src/std/debug.c:349:22: warning: pointer targets in passing argument 2 of ‘hl_debug_read’ differ in signedness [-Wpointer-sign]
>    hl_debug_read(pid, addr + (-((int_val)r)-1), (vbyte*)&out, sizeof(void*));
>                       ^~~~
> src/std/debug.c:104:13: note: expected ‘vbyte * {aka unsigned char *}’ but argument is of type ‘char *’
>  HL_API bool hl_debug_read( int pid, vbyte *addr, vbyte *buffer, int size ) {
>              ^~~~~~~~~~~~~
> make: *** [src/std/debug.o] Error 1
> Makefile:224: recipe for target 'src/std/debug.o' failed
> ```
> 
> Are 32 bit builds not supported anymore or am I missing something in the configuration?

This is the commit that broke 32 bit builds: https://github.com/HaxeFoundation/hashlink/commit/c516d990fa2973c56a8cf45e799079b94580ddfe.
In my opinion it's fine to drop 32 bit support, many distributions have dropped it already, and the vast majority of users should have 64-bit hardware by now.

Aurel300:
Looking at the commit you referenced @Apprentice-Alchemist, it might be enough to move that `case 11` into the `#ifdef HL_64` (and maybe `case 11: return -1;` in the `#else` branch). It tries to refer to XMM registers which I think only exist on a 64-bit architecture. [Here](https://elixir.bootlin.com/glibc/glibc-2.5/source/sysdeps/unix/sysv/linux/i386/sys/user.h#L26) is a 32-bit definition of `user_fpregs_struct` and [here](https://elixir.bootlin.com/glibc/glibc-2.5/source/sysdeps/unix/sysv/linux/x86_64/sys/user.h#L30) is a 64-bit one.

sebthom:
@Aurel300 can you merge this PR? It may not yet solve all issues but at least restores windows builds.

Simn:
Looks like the HL team was somehow misconfigured, you should have write access now.

Issue #477, opened 2021-10-01
`Std.parseInt` parses numbers with leading whitespaces incorrectly


kiwwisk:
## How to reproduce the error?

https://try.haxe.org/#B68880c8

If there's newline character at the beginning, `Std.parseInt` fails to parse the number correctly.

```hx
    trace(Std.parseInt("\n10"));
    trace(Std.parseInt("\n0x10"));
    trace(Std.parseInt("\r10"));
    trace(Std.parseInt("\r0x10"));
```

Prints:

```none
Test.hx:3: null
Test.hx:4: null
Test.hx:5: null
Test.hx:6: null
```

Should print:

```none
Test.hx:3: 10
Test.hx:4: 16
Test.hx:5: 10
Test.hx:6: 16
```

Note: The neko target also fails to parse the numbers correctly.

Yanrishatum:
Related: https://github.com/HaxeFoundation/hashlink/issues/330
Fix only checks for ` `  character, but not the CR / LF characters which are also considered to be white-space, meaning that described behavior of leading white-space being stripped is not respected.

ncannasse:
Having leading spaces in the parsed String in unspecified according to API documentation https://api.haxe.org/Std.html#parseInt

Yanrishatum:
@ncannasse Quite the opposite thought? Behavior description is explicit about leading whitespace.

> Leading whitespaces are ignored.

Common "whitespace" characters are space and newline characters.
Another related issue that touches up on all languages: https://github.com/HaxeFoundation/haxe/issues/8970
Note that fix for other languages was doing `(c > 8 && c < 14) || c == 32)` check, while HL checks for `32`, i.e. space character only.

See implementation: 
https://github.com/HaxeFoundation/hashlink/blob/master/src/std/bytes.c#L193-L196

And while for `parseFloat` it is not explicitly stated that leading whitespace is trimmed, it is a fairly common expectation, and again, on HL side it does perform stripping of leading spaces, but not whitespaces: 
https://github.com/HaxeFoundation/hashlink/blob/master/src/std/bytes.c#L183

Yanrishatum:
@ncannasse This solution has 2 flaws (one of which is critical):
[...]

PR #532, opened 2022-04-02
Fix parse_int with hexadecimal notation

tobil4sk:
Allow all whitespaces and + in front of hexadecimal number. Also, allows trailing text after hexadecimal number, like after normal number.

Related to #477.

### DIFF
--- a/src/hl.h
+++ b/src/hl.h
@@ -224,7 +224,7 @@ typedef wchar_t	uchar;
 #	define ustrlen		wcslen
 #	define ustrdup		_wcsdup
 HL_API int uvszprintf( uchar *out, int out_size, const uchar *fmt, va_list arglist );
-#	define _utod(s,end)	wcstod(s,end)
+#	define utod(s,len,end)		wcstod(s,end)
 #	define utoi(s,len,end)		wcstol(s,end,10)
 #	define ucmp(a,b)	wcscmp(a,b)
 #	define utostr(out,size,str) wcstombs(out,str,size)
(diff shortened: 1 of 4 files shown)


src/hl.h
```diff
@@ -224,8 +224,8 @@ typedef wchar_t	uchar;
 #	define ustrlen		wcslen
 #	define ustrdup		_wcsdup
 HL_API int uvszprintf( uchar *out, int out_size, const uchar *fmt, va_list arglist );
-#	define _utod(s,end)	wcstod(s,end)
-#	define _utoi(s,end)	wcstol(s,end,10)
+#	define utod(s,len,end)		wcstod(s,end)
+#	define utoi(s,len,end)		wcstol(s,end,10)
```
Simn, line 228: Why are we changing defines here?

tobil4sk: `_utod` and `_utoi` were used in actual `utod` and `utoi` functions which wrapped them and did an extra check ([see changes at the top of ucs2.c](https://github.com/HaxeFoundation/hashlink/pull/532/files#diff-b0ff64e049f3f8451842a16be609465714182f99675ae0c04b9d2940b0f6740cL29-L41)), however this check is no longer needed. Now they are direct replacements to get the function names the same between Windows and other systems, which handle strings differently. They were only used in one place, and this is how it was before 8d7b29c1865.

And the `len` was added because an extra parameter is needed in the non Windows version.

tobil4sk: Actually, sorry, the `len` should be removed from both versions as in the end it's not actually used.

tobil4sk:
@Simn Was there anything else left to do here?

src/std/ucs2.c
```diff
@@ -83,45 +67,47 @@ uchar *ustrdup( const uchar *str ) {
...
 
-int utoi( const uchar *str, uchar **end ) {
+int utoi( const uchar *str, const uchar **end ) {
 	char buf[17];
 	char *bend;
+	int result;
 	int i = 0;
-	int v;
-	while( is_space_char(*str) ) str++;
+	uchar sign = str[0];
```
Simn, line 92: Shouldn't this account for an empty string, in which case `str[0]` is out-of-bounds?

tobil4sk: Strings are null terminated, so an empty string would be `"\0"`, in that case `str[0]` would be `'\0'`. It's the same as `*str` which was done before.

Simn: Oh right, and I feel like you told me that before, haha...

PR #545, opened 2022-05-08
Fix make error related to pcre

tobil4sk:
The wrong variable was being used when building the regex part of the
standard library which was preventing compilation without a global
installation.

Only required if #515 is not merged first.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ HLFLAGS = -ldl
 LIBEXT = so
 LIBTURBOJPEG = -lturbojpeg
 
-PCRE_INCLUDE = -I include/pcre
+PCRE_FLAGS = -I include/pcre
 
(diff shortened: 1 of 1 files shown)


ncannasse:
Please contribute to the PR instead so everything can be merged at once.

tobil4sk:
The other PR fixes this as well, but I thought it might be good to have this as a separate PR so that the current build works even if it takes a while for the other one to be merged.

tobil4sk:
@ncannasse A user came across this issue recently on the Haxe discord. This PR simply fixes a typo introduced in #520, and it would be useful to have by itself unless the PCRE2 PR is likely to be merged any time soon. Please could you merge this?

tobil4sk:
Thanks!

PR #548, opened 2022-05-15
Fix CI

tobil4sk:


ncannasse:
Still some failures there ;)

tobil4sk:
Yes, a long list of weird compiler errors, eg:
```
C:\Program Files (x86)\Windows Kits\10\Include\10.0.17763.0\um\oaidl.h(487,17): error C2059: syntax error: '/' [D:\a\1\s\libhl.vcxproj]
C:\Program Files (x86)\Windows Kits\10\Include\10.0.17763.0\um\oaidl.h(502,17): warning C5103: pasting '/' and '/' does not result in a valid preprocessing token [D:\a\1\s\libhl.vcxproj]
C:\Program Files (x86)\Windows Kits\10\Include\10.0.17763.0\shared\wtypes.h(745): message : in expansion of macro '_VARIANT_BOOL' [D:\a\1\s\libhl.vcxproj]
C:\Program Files (x86)\Windows Kits\10\Include\10.0.17763.0\um\oaidl.h(502,17): error C2059: syntax error: '/' [D:\a\1\s\libhl.vcxproj]
C:\Program Files (x86)\Windows Kits\10\Include\10.0.17763.0\um\oaidl.h(529,40): warning C4094: untagged 'struct' declared no symbols [D:\a\1\s\libhl.vcxproj]
C:\Program Files (x86)\Windows Kits\10\Include\10.0.17763.0\um\oaidl.h(530,17): error C2059: syntax error: '}' [D:\a\1\s\libhl.vcxproj]
C:\Program Files (x86)\Windows Kits\10\Include\10.0.17763.0\um\oaidl.h(531,13): error C2059: syntax error: '}' [D:\a\1\s\libhl.vcxproj]
C:\Program Files (x86)\Windows Kits\10\Include\10.0.17763.0\um\oaidl.h(533,9): error C2059: syntax error: '}' [D:\a\1\s\libhl.vcxproj]
C:\Program Files (x86)\Windows Kits\10\Include\10.0.17763.0\um\oaidl.h(534,5): error C2059: syntax error: '}' [D:\a\1\s\libhl.vcxproj]
```
Not sure why there would be syntax errors, or why it's using the windows 10 kit and not 8.1 which is meant to be the one that is set up?

tobil4sk:
So there were some issues with how the SDK was configured in CMake, which prevented it from the correct one being uses (it has to be set before the project() call, at which point MSVC is not set so an if block I wrapped it in wasn't running).

Additionally, the compiler errors were still present even with the correct SDK. The 8.1 SDK (along with some older Windows 10 SDKs) requires the C language standard to be set to "Legacy MSVC", as some of the SDK header files depend on non-standard features. [This blog goes into more depth](https://blog.jetbrains.com/rscpp/2015/10/13/microsoft-cpp-compiler-bugs-ambiguities/#Eliminated_Types). Legacy MSVC is what is used by default, and it is what is configured by default for the projects in hl.sln.

tobil4sk:
@ncannasse The only thing that's left now is that DirectX 12 is not supported on Windows 8.1, so compilation is failing now. WIth DirectX12 support being added, it may be a good time to start building for Windows 10 instead of 8.1?

ncannasse:
I think that's what we did for latest vcxproj : upgraded to vs2019 + win10
sdk

tobil4sk:
Right now all the vcxproj files still have the windows 8.1 sdk, apart from the dx12 one:

For example:
https://github.com/HaxeFoundation/hashlink/blob/9cadbe5aacd32c22699a12b0a927627279a53506/hl.vcxproj#L33
https://github.com/HaxeFoundation/hashlink/blob/9cadbe5aacd32c22699a12b0a927627279a53506/libhl.vcxproj#L33

So the intention was to update to sdk 10 then?

tobil4sk:
Right now all the vcxproj files still have the windows 8.1 sdk, apart from the dx12 one:

For example:
https://github.com/HaxeFoundation/hashlink/blob/9cadbe5aacd32c22699a12b0a927627279a53506/hl.vcxproj#L33
https://github.com/HaxeFoundation/hashlink/blob/9cadbe5aacd32c22699a12b0a927627279a53506/libhl.vcxproj#L33

So the intention was to update to sdk 10 then?

tobil4sk:
@ncannasse Upgraded everything to win 10 sdk properly, and everything's building properly now so we're ready to merge. 👍

ncannasse:
Thanks !

PR #559, opened 2022-08-31
Allow multiple file selection in ui_choose_file

nspitko:
This is one of two PRs (The other in haxe for ui.hx) to add support for multiple selection in hl.UI.chooseFile.

Both do weird things for dumb windows API reasons, so I'll explain my sins in advance. Documentation for this API can be found here: https://docs.microsoft.com/en-us/windows/win32/api/commdlg/ns-commdlg-openfilenamea (We use the widechar version, but otherwise the call pattern is identical).

To enable multi select, we really only need to set a flag. Since we just pass a simple `Dynamic` across the fence for this, I just key off a new field and set the related flag. Note we also have to set `OFN_EXPLORER` else you get the win 3.x era version of this dialogue... which is magical in its own right.

The elephant in the room though is the larger buffer, and the fact that we return the entire thing. The buffer size increase is because this API doesn't really have any mechanism for letting you ask for how many bytes you need; hence we should always oversize. And the size chosen is simply due to that being the largest buffer you can even send this API. We'll need to use a newer API if we want anything more sophisticated. 

The reason we pass the whole buffer, instead of the old behavior of just copying the length, is because the return format is null terminated. So `wcslen` will only ever return the path. The haxe commit has the untangling code for converting this data into haxe strings. This code isn't performance critical, so the extra copy shouldn't be a huge deal, and `String.fromUCS2` doesn't rely on buffer size.

### DIFF
--- a/libs/ui/ui_win.c
+++ b/libs/ui/ui_win.c
@@ -281,9 +281,11 @@ HL_PRIM vbyte *HL_NAME(ui_choose_file)( bool forSave, vdynamic *options ) {
 	wref *win = (wref*)hl_dyn_getp(options,hl_hash_utf8("window"), &hlt_abstract);
 	varray *filters = (varray*)hl_dyn_getp(options,hl_hash_utf8("filters"),&hlt_array);
 	wchar_t *fileName = (wchar_t*)hl_dyn_getp(options,hl_hash_utf8("fileName"),&hlt_bytes);
+	bool multiple = hl_dyn_geti(options,hl_hash_utf8("multiple"),&hlt_bool);
+
 	OPENFILENAME op;
-	wchar_t filterStr[1024];
-	wchar_t outputFile[1024] = {0};
+	wchar_t filterStr[2048];
+	wchar_t outputFile[2048] = {0};
 	ZeroMemory(&op, sizeof(op));
 	op.lStructSize = sizeof(op);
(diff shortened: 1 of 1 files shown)


ncannasse:
Hi,
Looks fine, the only issue I have is that the windows specific implementation (in particular the 2048 buffer size) is visible in the Haxe code (idx < 2048). If we can get rid of that so that another platform API can return a valid buffer with no 2048 size limit I think we're good and it can be merged.
Best,
Nicolas

nspitko:
Pushed an update to both PRs.

That line was originally just a safety valve during testing, but removing and testing revealed a rather exciting bug: `nMaxFile` in the windows API is a *character* limit, not a buffer size! Since we're using the W version, the old value was actually a potential crash, and we just never hit it because of the windows path limit.

I fixed this and reverted some other unnecessary changes I made in the process. Buffer sizing is now accounting for wchar size, and appears to behave correctly when fed too many files (returns null)

ncannasse:
The final hl_copy_bytes value seems incorrect. maybe should be sizeof(outputFile) ?

nspitko:
Yep, that's clearly a bug. Fixed.

MSGhero:
I wonder if this helps the `File.browse()` issue I ran into https://github.com/HeapsIO/heaps/issues/783

PR #560, opened 2022-09-04
skip spaces during native signature comparison

CarlOlson:
The default code formatter for VSCode C++ will introduce a bug:
```c++
    // This
    DEFINE_PRIM(_ABSTRACT(state*), init, _NO_ARGS);
    // becomes this
    DEFINE_PRIM(_ABSTRACT(state *), init, _NO_ARGS);
```

```
src/module.c(513) : FATAL ERROR : Invalid signature for function hlcef@shutdown : PXstate*__v required but PXstate *__v found in hdll
```

This isn't a problem with hashlink, but it would be good to handle spaces.

BTW is there any way to add tests? It looks like CI only builds the project, right?

### DIFF
--- a/src/module.c
+++ b/src/module.c
@@ -509,8 +509,14 @@ static void hl_module_init_natives( hl_module *m ) {
 		p = tmp;
 		append_type(&p,n->t);
 		*p++ = 0;
-		if( memcmp(sign,tmp,strlen(sign)+1) != 0 )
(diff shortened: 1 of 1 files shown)


ncannasse:
I think you're using the abstract type wrong, it should not be a C type but a type name that does not contain any special character such as * (abstracts will always be pointers anyway).

PR #571, opened 2022-11-03
[make] Set rpath for hdll builds

tobil4sk:
Technically this shouldn't have been an issue anyway, but homebrew does a check to make sure all binaries that are linked to dynamic libraries can definitely load their linked libraries. Since no rpath was set for the hdlls, they were causing these tests to fail.

https://github.com/Homebrew/homebrew-core/pull/99237

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@ CFLAGS = -Wall -O3 -I src -msse2 -mfpmath=sse -std=c11 -D LIBHL_EXPORTS
 LFLAGS = -L. -lhl
 EXTRA_LFLAGS ?=
 LIBFLAGS =
+HDLLFLAGS = -L. -lhl
 HLFLAGS = -ldl
 LIBEXT = so
 LIBTURBOJPEG = -lturbojpeg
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -107,6 +108,7 @@ else
 # Linux
 CFLAGS += -m$(MARCH) -fPIC -pthread -fno-omit-frame-pointer
 LFLAGS += -lm -Wl,-rpath,.:'$$ORIGIN':$(INSTALL_LIB_DIR) -Wl,--export-dynamic -Wl,--no-undefined
+HDLLFLAGS += -Wl,-rpath,$(INSTALL_LIB_DIR)
```
Aurel300, line 111: It should be noted that if HashLink is meant to be shipped as part of a game (as in Shiro games), then the HDLLs should have an rpath that is something like `@executable_path/foo.hdll` (depends on the concrete directory structure of course). So the HDLLs will need to be patched with `install_name_tool`.

tobil4sk: These changes only affect Linux, right now we don't do anything rpath related on mac.

tobil4sk: These rpath issues have been a pain for a while. Maybe I'll open a separate issue so that we can discuss resolving them properly for all use cases.

tobil4sk: For now, I've added ORIGIN to the hdll's rpath so that it can load libhl from the same directory as well. This should cover the case where all the binaries are packaged in a single folder.

> then the HDLLs should have an rpath that is something like @executable_path/foo.hdll

Also note, this is about how the loader resolves libhl as a dependency of a hdll, not the other way around.

PR #574, opened 2022-11-21
Fix printing i64's.

Apprentice-Alchemist:


src/std/ucs2.c
```diff
@@ -208,7 +208,11 @@ HL_PRIM int uvszprintf( uchar *out, int out_size, const uchar *fmt, va_list argl
 					switch( c ) {
 					case 'd':
 						cfmt[i++] = 0;
-						size = sprintf(tmp,cfmt,va_arg(arglist,int));
+						if( cfmt[i-3] == 'l' ) {
```
Simn, line 211: Is that `i-3` indexing safe here?

PR #575, opened 2022-11-30
Added a print_flags to redirect the output to a file

Azrou:


ncannasse:
Since hl_to_utf8 can allocate memory, it should be done outside of
hl_blocking(true) ... hl_blocking(false) block.

PR #582, opened 2023-01-31
Fixed multiple warnings.

vitimiti:
Added explicit casts and added some NULL pointer checks to remove warnings that happen during compilation and prevent undefined behaviour.


libs/mysql/mysql.c
```diff
@@ -229,9 +229,9 @@ static CONV convert_type( enum enum_field_types t, int flags, unsigned int lengt
 	case FIELD_TYPE_DECIMAL:
 	case FIELD_TYPE_FLOAT:
 	case FIELD_TYPE_DOUBLE:
-	case 246: // 5.0 MYSQL_NEW_DECIMAL
+	case FIELD_TYPE_NEWDECIMAL: // 5.0 MYSQL_NEW_DECIMAL
 		return CONV_FLOAT;
-	case 245: // JSON
+	case FIELD_TYPE_JSON: // JSON
```
Aurel300, line 234: The comments can now be removed.

src/std/ucs2.c
```diff
@@ -80,7 +80,8 @@ double utod( const uchar *str, uchar **end ) {
 	}
 	buf[i] = 0;
 	result = strtod(buf,&bend);
-	*end = str + (bend - buf);
+	uchar *strcpy = ustrdup(str);
```
Aurel300, line 83: Best avoid calling this after an existing C function?

vitimiti: Yes, true, completely forgot about it.

src/std/ucs2.c
```diff
@@ -101,7 +102,8 @@ int utoi( const uchar *str, uchar **end ) {
 	}
 	buf[i] = 0;
 	result = strtol(buf,&bend,10);
-	*end = str + (bend - buf);
+	uchar *strcpy = ustrdup(str);
```
Aurel300, line 105: Here as well :)

vitimiti: Yes, I saw it just a bit after hitting Sync. Fixed. ^.^`

src/gc.c
```diff
@@ -878,7 +878,9 @@ void hl_free( hl_alloc *a ) {
 		size = (int)(b->p + b->size - ((unsigned char*)b));
 		prev = (int_val)b;
 		free(b);
-		b = n;
+		if (n) {
+			b = n;
+		}
```
Aurel300, line 883: This change seems wrong, and might be why CI is failing. This `while` is walking down a linked list, so it must terminate when `n == b->next == NULL`. If you avoid setting `b` for the final iteration, then `b` will always be the last node, so the loop will run forever; it is in fact stopped first because `free` rightly complains that the same `b` is `free`d again.

vitimiti: Hmmmm, makes sense, I'll see if I find a way to fix the warnings without breaking the CI.

vitimiti: I cannot think of how to fix this warning or the one with the i64 structure in src/std/bytes.c. So for now I am going to leave my fixes down to this point, as long as I haven't broken anything else, which I should the fix.

vitimiti:
It looks like my C++ training got in the way of how to properly initialize arrays to all zeros. I have added the proper C way and I will see if the tests now pass in MSVC.

PR #597, opened 2023-04-04
[CI] Go back to ubuntu 20.04

kLabz:
Ubuntu 18.04 is not available anymore on github actions.
I suppose we could rewrite it all using docker images, but would that be worth it?

As far as I understand, this means:
 * no more 32bit linux releases; support for which seem to have been dropped at most places
 * ubuntu 18 (which EOL is in a month or so) users will need to build from source or use an older release
 * anything else in a worse state than last month with ubuntu 18?

cc @Apprentice-Alchemist @tobil4sk  

Past work that this pretty much cancels: #529 #533

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,7 +27,7 @@ jobs:
 
         include:
           - target: linux
-            runner: ubuntu-18.04
+            runner: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


kLabz:
I don't know but I kinda doubt it since last version is indeed recent and there haven't been much changes related to hl recently

PR #608, opened 2023-07-19
Map add resize down when nentries becomes small

yuxiaomao:
This commit will resize down the map when its n < max*0.75, and reduce the remove execution time for a big but empty map.

### DIFF
--- a/src/std/maps.h
+++ b/src/std/maps.h
@@ -25,6 +25,7 @@ typedef struct {
 	int ncells;
 	int nentries;
 	int maxentries;
+	int minentries;
 } t_map;
 
 #ifndef _MNO_EXPORTS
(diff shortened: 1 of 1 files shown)


yuxiaomao:
I just implemented the Robin Hood Hashing for Hashmap. (References : https://www.youtube.com/watch?v=4cTKiTQtS_k&t=893s , https://programming.guide/robin-hood-hashing.html)
Tested with the previously described script, it seems to greatly reduce remove time without scarifying set/get time, however it occupied 50% more memory space.
- Removing all elements one by one for N= 500 000 only takes 0.029secs on my machine (compare to 0.041secs on JS, and 4.38secs with existing implementation, 0.772secs with previous optimization).
- Removing all elements one by one for N= 5 000 000 only takes 0.453secs 
- Get/Set time seems to be reduced 5% to 20% in most cases

The overall performance is now comparable to javascript target - even faster in some situations.

yuxiaomao:
TODO: after we're sure of this change, we'll need memdump + debugger support

PR #612, opened 2023-08-08
Remove deprecated use of @:enum abstract

qlambert-pro:
Although the changes are straight forward, I didn't manage to test them properly, because I didn't figure out how to add the repo as the git version to use for hlsdl in particular.


Simn:
CI is being weird here...

PR #613, opened 2023-08-18
Add support for Xbox Series: new macro HL_XBS, directx adapt

yuxiaomao:
This commit add a new target platform to hashlink : Xbox Series (console), aka Scarlett. Defined with macro `HL_XBS`. It does not contain Xbox specific features, except for some adaptation in DirectX12, but will allow the use of hashlink (instead of hashlink C) on the target platform.

Use with `-lib heaps`, add `-lib directx -D dx12 -D usesys` and definitions of `haxe.AudioTypes`, `haxe.GameController`, `haxe.System`.

- Tested on game with GDK version 230303, where 2D display are ok with Heaps.io
- Known issues:
  - Support only 1920x1080 resolution + 60 FPS, but not yet 2k / 4k / 120 FPS.
  - The `dx12.getRequiredIntermediateSize` return size+offset for HL_XBS, which is a tmp fix for `dx12.updateSubResource` error when `first != 0`


ncannasse:
I would prefer to avoid adding a specific configuration in vcproj for Xbox.
Instead we should use template based project generation as we have for
switch (in the private repo). Could you resubmit the PR with only C changes
?

Le ven. 18 août 2023 à 12:47, Yuxiao Mao ***@***.***> a
écrit :

PR #624, opened 2023-09-23
Restore basic debugging support on Linux.

Apprentice-Alchemist:
It doesn't work perfectly yet (for example stepping in/out is still broken), but at least breakpoints and breaking on exceptions works again.

### DIFF
--- a/src/debugger.c
+++ b/src/debugger.c
@@ -39,8 +39,12 @@ HL_API int hl_closure_stack_capture;
 
 static hl_socket *debug_socket = NULL;
 static hl_socket *client_socket = NULL;
-static bool debugger_connected = false;
-static bool debugger_stopped = false;
+#ifdef HL_LINUX
+extern int hl_debugger_connected;
+#else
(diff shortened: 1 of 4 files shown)


ncannasse:
That seems overly complex. Why allowing several debuggers to connect ?

rcstuber:
I think we should reopen and discuss this PR!
It looks valid for future Hashlink on ARM which is currently a work-in-progress by @Apprentice-Alchemist

Jarrio:
> That seems overly complex. Why allowing several debuggers to connect ?

> It's been a while since I wrote the code, but iirc the whole background thread and condition variable dance is needed because there's no way (that I found) to `waitpid` with a timeout.

Apprentice-Alchemist:
> > That seems overly complex. Why allowing several debuggers to connect ?
> 
> > It's been a while since I wrote the code, but iirc the whole background thread and condition variable dance is needed because there's no way (that I found) to `waitpid` with a timeout.

@Jarrio I was talking about my new implementation, not this one.

PR #639, opened 2023-12-16
Make macOS CI green again.

Apprentice-Alchemist:
- mbedtls@2 is keg-only, so it needs to be explicitly force-linked
- Switch to macOS 12, Homebrew doesn't support 11 anymore
   - Remove all pre-installed homebrew packages to prevent any issues during brew bundle.
   (This is probably overkill, but I do not have a better fix for now. Will send a follow-up PR if I find a better solution.)


Simn:
Thanks!

PR #644, opened 2024-01-17
Disable certain OpenGL functions on Apple platforms.

Apprentice-Alchemist:
Apple platforms only support up to OpenGL 4.1.


Simn:
Thanks!

ncannasse:
Could you throw exceptions instead of doing nothing? This would help with debugging.

Simn:
Oh, my bad for merging so quickly, I saw green CI and couldn't resist!

PR #647, opened 2024-01-31
Update actions used in CI workflow.

Apprentice-Alchemist:
Fixes the warnings about node.js 16 actions being deprecated.

Closes #640
Closes #641

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -83,7 +83,7 @@ jobs:
 
     - name: Add msbuild to PATH
       if: matrix.build_system == 'vs2019'
-      uses: microsoft/setup-msbuild@v1
+      uses: microsoft/setup-msbuild@v2
       with:
         vs-version: '[16.0,17.0)'
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -287,8 +287,9 @@ jobs:
 
 
     - name: "Share: build artifact"
-      uses: actions/upload-artifact@v3
+      uses: actions/upload-artifact@v4
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -342,7 +330,7 @@ jobs:
...
           --prerelease \
           --title "HashLink Nightly Build" \
-          "artifact/hashlink-${short_commit}-darwin.tar.gz#hashlink-latest-darwin.tar.gz" \
-          "artifact/hashlink-${short_commit}-linux-amd64.tar.gz#hashlink-latest-linux-amd64.tar.gz" \
-          "artifact/hashlink-${short_commit}-win32.zip#hashlink-latest-win32.zip" \
-          "artifact/hashlink-${short_commit}-win64.zip#hashlink-latest-win64.zip"
+          "darwin-make-64/hashlink-${short_commit}-darwin.tar.gz#hashlink-latest-darwin.tar.gz" \
+          "linux-make-64/hashlink-${short_commit}-linux-amd64.tar.gz#hashlink-latest-linux-amd64.tar.gz" \
+          "windows-vs2019-32/hashlink-${short_commit}-win32.zip#hashlink-latest-win32.zip" \
+          "windows-vs2019-64/hashlink-${short_commit}-win64.zip#hashlink-latest-win64.zip"
```
Simn, line 336: Why are we changing this?

Apprentice-Alchemist: In v4 of actions/upload-artifact CI artifacts have become immutable, so each matrix combination now uploads a uniquely named artifact instead of adding the zip/tar to the existing artifact.
The artifacts used for the release are still the same as before.

Issue #656, opened 2024-02-21
Get live objects from a specific type


SpiraTom:
With the aim of improving memory leak detection, I need to get occurences of live objects from a specific type in runtime.

Is it possible to implement a method to get live objects `varray *hl_get_live_objects( hl_type *t, int maxCount )` ?

PR #657, opened 2024-02-28
Add get live objects

yuxiaomao:
For https://github.com/HaxeFoundation/hashlink/issues/656 , usage (API should be in haxe someday with hl_ver >= 1.15.0):

```haxe
class MemoryTools {
	/**
		Count live objects of a given class and get at most `maxCount` elements in an array.
	**/	
	public static function getLiveObjects(cl : Class<Dynamic>, maxCount : Int = 0) {
		var arr = new hl.NativeArray<Dynamic>(maxCount);
		var count = _getLiveObjects(cast(cl,hl.BaseType).__type__, arr);
		var objs = new Array<Dynamic>();
		for (i in 0...maxCount) {
			if (arr[i] == null) break;
			objs.push(arr[i]);
		}
		return { count : count, objs : objs };
	}

	@:hlNative("std", "gc_get_live_objects")
	static function _getLiveObjects(type : hl.Type, arr : hl.NativeArray<Dynamic>) : Int { return 0; }
}
```

Then,

```haxe
MemoryTools.getLiveObjects(h2d.col.Point.PointImpl, 5);
MemoryTools.getLiveObjects(std.Type.resolveClass("h2d.col.PointImpl"), 5);
```

However it count only exactly the same type (no child class instance, no abstract).

### DIFF
--- a/src/gc.c
+++ b/src/gc.c
@@ -1445,6 +1445,33 @@ HL_API void hl_gc_dump_memory( const char *filename ) {
 	gc_global_lock(false);
 }
 
+static int live_obj_count;
+static hl_type *live_obj_t;
+
+static void gc_count_live_block(void *block, int size) {
+	hl_type *t = *(hl_type **)block;
+	if (t == live_obj_t)
+		live_obj_count += 1;
+}
+
+static void gc_count_live_page(gc_pheader *p, int private_data) {
+	gc_iter_live_blocks(p, gc_count_live_block);
+}
+
+static int hl_gc_count_live_objects(hl_type *t) {
+	gc_global_lock(true);
+	gc_stop_world(true);
+	gc_mark();
+
+	live_obj_count = 0;
+	live_obj_t = t;
+	gc_iter_pages(gc_count_live_page);
+
+	gc_stop_world(false);
+	gc_global_lock(false);
+	return live_obj_count;
+}
+
 #ifdef HL_VCC
 #	pragma optimize( "", off )
 #endif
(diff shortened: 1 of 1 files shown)


ncannasse:
Hi,

I think you should restrict the counting to pages which are MEM_DYNAMIC (or
something similar)


Le mer. 28 févr. 2024 à 15:09, Yuxiao Mao ***@***.***> a
écrit :

yuxiaomao:
I tried restrict to not `MEM_KIND_NOPTR`, it fail to find object that does not have pointer inside such as `h2d.col.PointImpl`, which behave even more differently than the memdump tool.

yuxiaomao:
I have edited the example usage, and I'm now satisfied with the code at hashlink side.

src/gc.c
```diff
@@ -1445,6 +1445,45 @@ HL_API void hl_gc_dump_memory( const char *filename ) {
...
 
+static int live_obj_count;
+static hl_type *live_obj_t;
+static int live_obj_page_kind;
+static varray *live_objs;
+static int live_objs_index;
+
+static void gc_count_live_block( void *block, int size ) {
+	hl_type *t = *(hl_type **)block;
+	if( t != live_obj_t || size < sizeof(void*) ) return;
```
ncannasse, line 1456: you should maybe check size before reading block?

src/gc.c
```diff
@@ -1445,6 +1445,45 @@ HL_API void hl_gc_dump_memory( const char *filename ) {
...
+	live_obj_count += 1;
```
ncannasse, line 1457: ++ for code style

src/gc.c
```diff
@@ -1445,6 +1445,45 @@ HL_API void hl_gc_dump_memory( const char *filename ) {
 	gc_global_lock(false);
 }
 
+static int live_obj_count;
+static hl_type *live_obj_t;
+static int live_obj_page_kind;
+static varray *live_objs;
+static int live_objs_index;
```
ncannasse, line 1452: maybe declare these in a struct, since they belong together.

src/gc.c
```diff
@@ -1445,6 +1445,45 @@ HL_API void hl_gc_dump_memory( const char *filename ) {
...
+	}
+}
+
+static void gc_count_live_page( gc_pheader *p, int private_data ) {
+	if( p->page_kind == live_obj_page_kind )
+		gc_iter_live_blocks(p, gc_count_live_block);
+}
+
+static int hl_gc_get_live_objects( hl_type *t, varray *arr ) {
+	if( t->kind != HOBJ ) return -1;
```
ncannasse, line 1470: maybe we could allow not HOBJ - but then we check all pages?

PR #661, opened 2024-03-09
upstream some patches for OpenBSD

rfht:
Fairly straightforward stuff, mostly adding OpenBSD to `#ifdef`s. Our pthread API is slightly different.

### DIFF
--- a/libs/mysql/socket.c
+++ b/libs/mysql/socket.c
@@ -74,7 +74,7 @@ void psock_init() {
 
 PSOCK psock_create() {
 	PSOCK s = socket(AF_INET,SOCK_STREAM,0);
-#	if defined(OS_MAC) || defined(OS_BSD)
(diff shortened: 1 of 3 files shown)


rfht:
any chance for this to go in?

rfht:
just checking if this looks okay...

tobil4sk:
It might be best to create new HL_OPENBSD and OS_OPENBSD defines and use those.

rfht:
I can do that if that's the preferred style; it will just be a define equivalent to `__OpenBSD__` but I see that it would make it easier to read and the constant names more consistent

yuxiaomao:
Please update with new `HL_OPENBSD` (in hl.h) and `OS_OPENBSD` (in libs/mysql/osdef.h) and use those definition.

Issue #658, opened 2024-03-02
JIT ERROR 0 mixing hl.UI8 and Float


ProfDiesel:
The following code triggers a `JIT ERROR 0 (jit.c line 554)` with hashlink 1.14.0 and haxe compiler 4.3.3.  

**Main.hx**
```
abstract A(hl.UI8) {
    public function f() {
        final _ = (cast this) * 0.0;
        return 0;
    }
}

class Main {
    public static function main():Void {
        trace(cast (0, A).f());
    }
}
```
Compile and run with: `haxe --main Main --hl main.hl && hl ./main.hl`
NB: changing the underlying type of the abstract for an `Int` gets around the issue. 

---
Narrowing down the problem from the original code, I came across a slightly different message with the code below :
`JIT ERROR 0 (jit.c line 968)`
Probably related.

```
abstract A(hl.UI8) {
    public function f() {
        this * 0.0;
    }
}

class Main {
    public static function main():Void {
    }
}
```

yuxiaomao:
Cannot reproduce with Haxe git + Hashlink git, the program complete without issues.

yuxiaomao:
Can repro with Linux, Ubuntu 22.04

Issue #142, opened 2018-05-02
Std.int(hl.UI8) causes jit error on linux


shakesoda:
Anything to do with int<->UI8 conversions seems to be blowing up on linux jit. this currently breaks jit for my game (c builds are fine).

repro:
```haxe
class Test {
	static function explode(v: hl.UI8) {
		trace(Std.int(v));
	}
	static function main() {
		explode(5);
	}
}
```
output: `JIT ERROR 0 (jit.c line 517)`

tested on ubuntu 16.04 and 18.04, latest commit of haxe and hl as of this writing.

ncannasse:
I'm not able to reproduce on both Win32 and Win64. Could you check again and give me the C stack trace so I know where this comes from?

shakesoda:
Here's the trace.

```op (ctx=ctx@entry=0x555555772dd0, o=o@entry=MOV8, a=a@entry=0x555555772ea0, 
    b=b@entry=0x555555772e88, mode64=mode64@entry=0 '\000') at src/jit.c:517
517				ASSERT(0);
(gdb) bt
#0  op (ctx=ctx@entry=0x555555772dd0, o=o@entry=MOV8, 
    a=a@entry=0x555555772ea0, b=b@entry=0x555555772e88, 
    mode64=mode64@entry=0 '\000') at src/jit.c:517
#1  0x000055555555b9b8 in op32 (ctx=ctx@entry=0x555555772dd0, o=o@entry=MOV8, 
    a=a@entry=0x555555772ea0, b=b@entry=0x555555772e88) at src/jit.c:791
#2  0x000055555555c5ca in copy (ctx=ctx@entry=0x555555772dd0, 
    to=to@entry=0x555555772ea0, from=0x555555772e88, size=1) at src/jit.c:1106
#3  0x000055555555c913 in copy_from (ctx=ctx@entry=0x555555772dd0, 
    to=to@entry=0x555555772ea0, from=from@entry=0x55555577e2d0)
    at src/jit.c:1238
#4  0x000055555556178c in hl_jit_function (ctx=ctx@entry=0x555555772dd0, 
    m=m@entry=0x55555577fbd0, f=f@entry=0x55555578c278) at src/jit.c:3021
#5  0x0000555555565aa2 in hl_module_init (m=0x55555577fbd0) at src/module.c:382
#6  0x0000555555564c8b in main (argc=0, argv=0x7fffffffde68) at src/main.c:156
```

[...]

PR #665, opened 2024-03-20
Fix UI8 assign on Linux

yuxiaomao:
Fixes https://github.com/HaxeFoundation/hashlink/issues/142 , fixes https://github.com/HaxeFoundation/hashlink/issues/658 for me without breaking our game on Linux (Ubuntu 22.04).

I'm not very sure what I'm doing, so here is some description:
I found that Linux & Windows have different set of CPU registers (if I use the same cpu set, I have segfault)
https://github.com/HaxeFoundation/hashlink/blob/8bc5f8f0d21c38805ca3457e5298a27d96c396ad/src/jit.c#L227-L239

Esi, Edi are used in Linux, however they are not 8 bit register (remove the id check on RCPU fixes the issues, but will make our game to exit "normally" without actually run and I don't understand why)
https://github.com/HaxeFoundation/hashlink/blob/8bc5f8f0d21c38805ca3457e5298a27d96c396ad/src/jit.c#L542-L544

The UI8 op failed during a MOV8 (called by copy-copyfrom), with a RCPU id = 1, b RCPU id = 7 (Edi)
https://github.com/HaxeFoundation/hashlink/blob/8bc5f8f0d21c38805ca3457e5298a27d96c396ad/src/jit.c#L552-L554

So I move the `!is_reg8(from)` also for the case when to.kind = RCPU. I also force the Reg choice to `RCPU_8BITS` because by the name it make sens, but it seems not necessary for the fix.

### DIFF
--- a/src/jit.c
+++ b/src/jit.c
@@ -1135,8 +1135,9 @@ static preg *copy( jit_ctx *ctx, preg *to, preg *from, int size ) {
 					op32(ctx,SHR,to,pconst(&p,24));
 					break;
 				}
-			} else if( !is_reg8(from) ) {
-				preg *r = alloc_reg(ctx, RCPU_CALL);				
+			}
+			if( !is_reg8(from) ) {
+				preg *r = alloc_reg(ctx, RCPU_8BITS);
 				op32(ctx, MOV, r, from);
 				RUNLOCK(r);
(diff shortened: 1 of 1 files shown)


ncannasse:
Seems good

PR #666, opened 2024-03-21
Check for Vulkan or Metal request and don't force OpenGL

onehundredfeet:
To make adding Vulcan or metal easier, I removed the forced OpenGL specification on the window creation and only add it if neither Vulcan nor metal is added.

This will make it backwards compatible while also allowing Vulkan or Metal windows to be created.


onehundredfeet:
The checks are failing because Neko not installing.

ncannasse:
Look at the vulkan branch if you want to see how I did the SDL setup here : https://github.com/HaxeFoundation/hashlink/tree/vulkan
And this in particular
https://github.com/HaxeFoundation/hashlink/commit/d15f466b59353dc3681df65b111f602c3139bafa

onehundredfeet:
It looks like the sdl_vulkan header is not available on the build machines.  What's the best solution here?

1) Guard using a define?
2) Add the headers to the repo?
3) Something else?

PR #672, opened 2024-04-01
Add support for SDL_GetGlobalMouseState

nspitko:
This allows us to track drag/resize movements that go outside the window; fixing the issue where dragging too fast will lose a handle on the window. The motivation behind this feature is to support ImGui multi-window viewports

Viewports will function without this, but moving/resizing windows sucks, as the mouse can move more than the title bar's width per frame, causing it to lose track of the window and "drop" it.

Unsure if there's a parallel to this in DX, but we can't use that backend currently anyway since it lacks single driver multi window support at the moment. Given this limitation I'm unsure there's any value in adding a similar function or common routing in hldx, but I can do this work if it's necessary for merge.

### DIFF
--- a/libs/sdl/sdl.c
+++ b/libs/sdl/sdl.c
@@ -474,6 +474,10 @@ HL_PRIM bool HL_NAME(get_window_grab)(SDL_Window* window) {
 	return SDL_GetWindowGrab(window);
 }
 
+HL_PRIM int HL_NAME(get_global_mouse_state)(int* x, int* y) {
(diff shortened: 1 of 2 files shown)


ncannasse:
Please add `@:hlNative("?sdl","get_global_mouse_state")` so someone using new headers without recompiling the library won't get a startup error if they're not using the function.

nspitko:
Oh neat, I didn't know you could do that.

Added.

PR #677, opened 2024-04-18
Fix multiple instances of undefined behaviour and crashes when using ubsan

GasInfinity:
While compiling with [zig cc](https://ziglang.org/learn/overview/#zig-is-also-a-c-compiler) I encountered crashes due to undefined behaviour (It enables ubsan by default in debug/safe builds). This PR aims to fix some code which causes ub

Some of the fixes include:
 * Avoid passing NULL to src or dest in memcpy()
 * Avoid incrementing/decrementing a NULL pointer
 * Avoid incrementing a pointer by 0
 * Avoid doing signed bit shifting by casting to unsigned first

### DIFF
--- a/src/gc.c
+++ b/src/gc.c
@@ -61,8 +61,8 @@
 //		0x0000000YXXX0000
 //		0x0007FY0YXXX0000
 static int_val gc_hash( void *ptr ) {
-	int_val v = (int_val)ptr;
-	return (v ^ ((v >> 33) << 28)) & 0x0000000FFFFFFFFF;
+	uint_val v = (uint_val)ptr;
+	return (int_val)((v ^ ((v >> 33) << 28)) & 0x0000000FFFFFFFFF);
 }
 #endif
 
@@ -271,8 +271,12 @@ HL_PRIM void hl_add_root( void *r ) {
 	if( gc_roots_count == gc_roots_max ) {
 		int nroots = gc_roots_max ? (gc_roots_max << 1) : 16;
 		void ***roots = (void***)malloc(sizeof(void*)*nroots);
-		memcpy(roots,gc_roots,sizeof(void*)*gc_roots_count);
-		free(gc_roots);
+
+		if(gc_roots) {
+			memcpy(roots,gc_roots,sizeof(void*)*gc_roots_count);
+			free(gc_roots);
+		}
+
 		gc_roots = roots;
 		gc_roots_max = nroots;
 	}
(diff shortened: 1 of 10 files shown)


src/gc.c
```diff
@@ -271,8 +271,12 @@ HL_PRIM void hl_add_root( void *r ) {
 	if( gc_roots_count == gc_roots_max ) {
 		int nroots = gc_roots_max ? (gc_roots_max << 1) : 16;
 		void ***roots = (void***)malloc(sizeof(void*)*nroots);
-		memcpy(roots,gc_roots,sizeof(void*)*gc_roots_count);
-		free(gc_roots);
+
+		if(gc_roots) {
```
Aurel300, line 275: I wonder: is this assumption ever violated? Can we use an `assume` intrinsic or similar?

src/std/bytes.c
```diff
@@ -143,7 +145,9 @@ HL_PRIM int hl_bytes_rfind( vbyte *where, int len, vbyte *which, int wlen ) {
 }
 
 HL_PRIM void hl_bytes_fill( vbyte *bytes, int pos, int len, int value ) {
-	memset(bytes+pos,value,len);
+	if(bytes) {
```
Aurel300, line 148: So this will now fail silently? If this check is performed, might as well emit a warning or something?

src/gc.c
GasInfinity: Well, when hashlink starts gc_roots == NULL.

```
frame #5: 0x00007ffff7e0abc7 libhl.so.1`hl_add_root(r=0x00007ffff7f1c1a0) at gc.c:274:3
   271          if( gc_roots_count == gc_roots_max ) {
   272                  int nroots = gc_roots_max ? (gc_roots_max << 1) : 16;
   273                  void ***roots = (void***)malloc(sizeof(void*)*nroots);
-> 274                  memcpy(roots,gc_roots,sizeof(void*)*gc_roots_count);
   275                  free(gc_roots);
   276                  gc_roots = roots;
   277                  gc_roots_max = nroots;
(lldb) v
(void *) r = 0x00007ffff7f1c1a0
(int) nroots = 16
(void ***) roots = 0x0000000001194c60
(lldb) print gc_roots
(void ***) 0x0000000000000000
```

So, we can add that check or initialize it before calling hl_add_root() somewhere else.

src/std/bytes.c
GasInfinity: That happened when I was testing the changes with some openfl templates. It seems that somewhere it passes NULL to `hl_bytes_fill` and `hl_bytes_blit`.

GasInfinity:
Closing due to lack of activity. If someone wants to get this merged, just @ me and I'll continue working on this!

PR #679, opened 2024-04-22
Fix debug Linux

yuxiaomao:
I made some changes for the hashlink-debugger vscode extension on Linux https://github.com/vshaxe/hashlink-debugger/compare/1.4.8...1.4.9 . They are not needed for the command line version (because you can't give command when the program is still in "running" mode).
As I also made a big hack for waitpid timeout (for the moment it works on my machine without need thread+semaphore), ping @ncannasse

### DIFF
--- a/src/std/debug.c
+++ b/src/std/debug.c
@@ -25,6 +25,7 @@
 #	include <sys/wait.h>
 #	include <sys/user.h>
 #	include <signal.h>
+#	include <errno.h>
 #	define USE_PTRACE
 #endif
 
@@ -84,6 +85,7 @@ HL_API bool hl_debug_stop( int pid ) {
 #	elif defined(MAC_DEBUG)
 	return mdbg_session_detach(pid);
 #	elif defined(USE_PTRACE)
+	kill(pid, SIGTRAP); // DETACH needs ptrace-stop
 	return ptrace(PTRACE_DETACH,pid,0,0) >= 0;
 #	else
 	return false;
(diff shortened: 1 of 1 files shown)


yuxiaomao:
My questions:
- Maybe `kill` in `hl_debug_stop` should be at Haxe side (in this case there is a need to either always stop before detaching, or detect Linux - I'm not sure which choice is better)?
- Is that ok to have a strange hack on `waipid` ? or maybe it's better to use a thread like for Mac ? or have a different behavior when running node and when running hashlink?

src/std/debug.c
```diff
@@ -84,6 +85,7 @@ HL_API bool hl_debug_stop( int pid ) {
 #	elif defined(MAC_DEBUG)
 	return mdbg_session_detach(pid);
 #	elif defined(USE_PTRACE)
+	kill(pid, SIGTRAP); // DETACH needs ptrace-stop
```
Apprentice-Alchemist, line 88: I think `SIGSTOP` would be more appropriate here.

src/std/debug.c
```diff
@@ -248,14 +250,21 @@ HL_API int hl_debug_wait( int pid, int *thread, int timeout ) {
...
 #	elif defined(USE_PTRACE)
 	int status;
-	int ret = waitpid(pid,&status,0);
-	//printf("WAITPID=%X %X\n",ret,status);
+	// *** HACK ***
+	// usleep here is needed for a good result.
+	// Without it, waitpid can miss many stop event;
+	// With it, and more we wait, less we miss stop event.
+	usleep(100 * 1000);
+	int ret = waitpid(pid, &status, WNOHANG);
```
Apprentice-Alchemist, line 258: I suspect that `waitpid` with `WNOHANG` is also vulnerable to a race condition with libuv.

I'm not sure how to fix this other than by doing the same thread & semaphore dance as the Mac debugger.

yuxiaomao: I also suspect the `waitpid` hack was for libuv (or vscode) waiting for all childs.

I tested your implementation with thread (it helps me a lot!), but it failed when we are actually waiting (by default the vscode extension call wait with a timeout = 0) and has a timeout. I didn't managed to get it works yet :'(
You said @ncannasse refused your PR due to complexity, so I tried this hack. I will do the thread & semaphore if hack is not a good option either.

Apprentice-Alchemist: That was another PR, I never actually submitted the thread & semaphore implementation.

yuxiaomao: Sorry I misunderstood :(

PR #680, opened 2024-04-23
allow casting between hl.Abstracts

fourst4r:
Not sure if should check `abs_name` here or not.
I mainly wanted this to be able to do a `hl_dyn_call` with an abstract as argument.


ncannasse:
There should definitely be a check for abs_name before we allow this. Usually the  `to == from` would already handle this

PR #681, opened 2024-05-04
SSL bugfixes.

Apprentice-Alchemist:
- get rid of the brew uninstall hack, it doesn't seem to be needed anymore
- only define MSG_NOSIGNAL if it is actually undefined
- Use `SecTrustCopyAnchorCertificates` to get root certs on macOS instead of the deprecated keychain apis.
- Initialize PSA crypto when it is present (in mbedtls 3.6 TLS 1.3 support is turned on by default which uses PSA crypto)

The partially fixes the macOS HTTPS failures encountered in https://github.com/HaxeFoundation/haxe/pull/11638.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -107,8 +107,6 @@ jobs:
             ;;
 
           darwin*)
-            brew list --formula | xargs brew uninstall --force --ignore-dependencies
-            brew list --cask | xargs brew uninstall --force --ignore-dependencies
             brew update
             brew bundle
             ;;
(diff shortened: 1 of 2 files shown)


Simn:
@yuxiaomao Could you check if this works for you guys? If so I'd like to see this merged because there are some related open Haxe issues.

yuxiaomao:
It's very Mac specific so I can't really understand it :x. But as I can't find any problem in the code and build on both CI / local machines seems fine, let's merge it.

PR #686, opened 2024-05-21
[mysql] Fix compilation.

Apprentice-Alchemist:
This fixes errors encountered when compiling Hashlink on Arch Linux and Fedora.


yuxiaomao:
I don't really understand but seems legit and CI is green.

Issue #685, opened 2024-05-19
JIT/VM calls incorrect method


m0rkeulv:
When testing one of my projects with HL i got this unexpected behavior where the VM seems to call the wrong function, it only happens when the object is retrieved from “container” object. I have created a small project that hopefully should be able to reproduce this problem.

it was compiled on `Windows 10` using `haxe 4.3.4` and running  on `Hashlink 1.14.0` (downloaded form the release  page here on Git)

Code to reproduce the problem: [HL_JIT_Bug.zip](https://github.com/HaxeFoundation/hashlink/files/15369570/HL_JIT_Bug.zip)

![image](https://github.com/HaxeFoundation/hashlink/assets/3193925/5e80e2d5-b9ca-47ef-b386-7d94613965f3)
![image](https://github.com/HaxeFoundation/hashlink/assets/3193925/1f577c62-c6dc-48f2-a5bc-8889c85dadfe)

Replacing the interface used  with a typedef pointing to the base class changes the behavior.
It looks like the correct method is called, but it causes an `Access Violation`, 
The debugger also no longer seems to recognize  the type showing it as `{...}`

code to reproduce:  [HL_JIT_Bug_typedef.zip](https://github.com/HaxeFoundation/hashlink/files/15369648/HL_JIT_Bug_typedef.zip)

![image](https://github.com/HaxeFoundation/hashlink/assets/3193925/951911dc-272f-42df-ba36-73cdf5f5ac28)
![image](https://github.com/HaxeFoundation/hashlink/assets/3193925/aeb14d3a-4e3d-4251-9211-4b380e690dcb)

If i change the container to specific type instead of  interface (replacing `User`with `BasicUser`) the code runs fine, so it seems to be related to the interface part.

UPDATE:  

After some more experimentation  i think i have found the cause of the issue, i think its when one interface inherits/ extends another one and they contain identical member names but returns a  type extending the original type. this works on other targets and compiles fine, so i guess this is an issue with the  HL VM

```haxe 
//Entity.hx (interface)
 function getVariable(varName:String):Variable
 //  User.hx (interface) (User extends Entity)
 function getVariable(name:String):UserVariable;
 ```

m0rkeulv:
i've  been playing around with the code and  i got things working, but i am not a hundred percent sure i am not breaking anything. my changes are in safe  cast, to me it looks like the issue is that  since my other class contains a member that is first in the fields array it breaks  and return false without checking if the expected fields are included.

[...]

PR #691, opened 2024-06-18
when doing safe_cast on HVIRTUAL, ignore extra fields from input type

m0rkeulv:
Not 100% sure this does not break anything else, but what i am attempting here is to solve the issue where interfaces that extends other interfaces and override a method signature by changing its return type ends up corrupting something making the runtime call a different unrelated method.

this fixes the interface-wrong method issue (https://github.com/HaxeFoundation/hashlink/issues/685).

Note that it does not fix the Access Violation you get from typedef example, however they seem to be related, if i on purpose set hl_vfields(v)[i]  to the method reference instead of NULL when hl_safe_cast returns false in  hl_to_virtual, the typedef variant runs just fine as well.

### DIFF
--- a/src/std/types.c
+++ b/src/std/types.c
@@ -191,15 +191,21 @@ HL_PRIM bool hl_safe_cast( hl_type *t, hl_type *to ) {
 		return false;
 	switch( t->kind ) {
 	case HVIRTUAL:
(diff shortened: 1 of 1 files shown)


m0rkeulv:
Problem fixed in Haxe https://github.com/HaxeFoundation/haxe/pull/11723

PR #697, opened 2024-07-13
Add set window focus support to hldx/hlsdl

nspitko:
This adds winRaise to hlSdl and winSetFocus to hldx. These are plumbed through to their respective window classes as well.

Both methods tested and working on windows 10

For winSetFocus (directx), minver is Windows 2000, not deprecated, should be safe.
for winRaise (sdl), min SDL version is 2.0.0, we should trust SDL to handle the rest.

Naming was chosen to match the underlying functions and following existing naming conventions, but I can rename either one for consistency if that is preferred.

This has a variety of minor uses in games, but the main driver is for imgui multi-window work, where we want a specific tool to grab focus for workflow reasons.

### DIFF
--- a/libs/directx/dx/Window.hx
+++ b/libs/directx/dx/Window.hx
@@ -145,6 +145,10 @@ class Window {
 		winResize(win, 2);
 	}
 
+	public function focus() {
+		winSetFocus(win);
+	}
+
 	public function getNextEvent( e : Event ) : Bool {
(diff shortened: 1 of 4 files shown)


yuxiaomao:
Hi, could you please confirm that this PR is still up-to-date?

nspitko:
Pulled upstream, still looks good to me.

PR #703, opened 2024-08-22
Fix gl.c for gcc 14.1.1

gogoprog:
I'm not familiar with `glMultiDrawElementsIndirectCountARB` so please confirm this fix is valid.

I had to do this cast for `gcc 14.1.1`.


kLabz:
This fix was needed for me too to fix hashlink compilation on linux :+1:

yuxiaomao:
Just checked with our graphics prog, seems good!

PR #704, opened 2024-09-01
Fix mingw errors in hlc_main.c

### LINKED ISSUE #687: Unable to build C output on Windows gcc
I have compiled some code from Haxe to C, but I am unable to build it on Windows using gcc. I am using gcc installed from MSYS2's mingw-w64-x86_64-gcc package.

This is my test code:
```haxe
class Main {
    static function main() {
        trace("Hello, world!");
    }
}
```

This is my build.hxml:
```hxml
--class-path src
--main Main
--hl out/hlc/main.c
```

And this is the error I get:
```
gcc -std=c11 -c out/hlc/main.c -o out/bin/test.o -I out/hlc -I/c/Users/nabca/Documents/src/hashlink/x64/Debug/../../src
In file included from out/hlc/main.c:6:
C:/Users/nabca/Documents/src/hashlink/src/hlc_main.c: In function 'WinMain':
C:/Users/nabca/Documents/src/hashlink/src/hlc_main.c:133:16: error: implicit declaration of function 'wmain' [-Wimplicit-function-declaration]
  133 |         return wmain(__argc, __argv);
      |                ^~~~~
```

I did notice that if I modify hlc_main.c so that the `WinMain` and `wWinMain` functions are defined after `wmain` and `main`, as well as change the parameters of `wmain` to `int argc, char *argv[]` instead of `int argc, uchar *argv[]`, the code will be able to compile successfully.

tobil4sk:
Closes #687

These issues were errors with GCC, but they were also warnings on MSVC.

### DIFF
--- a/src/hlc_main.c
+++ b/src/hlc_main.c
@@ -164,7 +164,7 @@ int main(int argc, char *argv[]) {
 
 #if defined(HL_WIN_DESKTOP) && !defined(_CONSOLE)
 int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdLine, int nCmdShow) {
-	return wmain(__argc, __argv);
+	return wmain(__argc, __wargv);
 }
 #elif defined(HL_XBS)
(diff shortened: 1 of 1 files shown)


tobil4sk:
Thanks!

PR #705, opened 2024-09-01
Support cmake installs with VS project templates

tobil4sk:
When hashlink is installed via cmake, the library files end up in the `lib` subdirectory


yuxiaomao:
Hi, I never used cmake, could you please detail a little bit how to use cmake for generation and what libraries ends up in `lib`? Especially, it that possible to just adjust where is the HASHLINK variables?
I know that I have made `ci` template, where I have to split src directory and dll directory, I'm wondering if that's a similar problem.

yuxiaomao:
Thank you for the explanation!
I just did a quick search on CMake and merged the other PR, as it can help CMake users x).
I see that CMake generate bin/include/lib structure for all targets, and this structure is different than what make/vs2019 generates. If possible, I think it would better to let them generate the same structure and keep things simple?

tobil4sk:
> I just did a quick search on CMake and merged the other PR, as it can help CMake users x).

Thanks!

> If possible, I think it would better to let them generate the same structure and keep things simple?

Ok, that sounds good. I'll open a new PR that does that instead!

tobil4sk:
I've opened: #709

PR #706, opened 2024-09-01
[hlc] Implement automatic compilation with make and Build.xml templates

tobil4sk:
This adds two new templates to the hashlink support library to simplify hlc compilation: `make` for a regular Makefile, and `hxcpp` for a Build.xml. Both of these implement automated compilation, so if you add either of these then an executable is generated automatically:

```hxml
-D hlgen.makefile=make
-D hlgen.makefile=hxcpp
```

The Makefile is smart so it will only recompile files whose dependencies have changed. I've tested it on Linux and on Windows with mingw.

Like with the Visual Studio templates, on Windows the templates expect a `HASHLINK` environment variable to point to the hashlink installation. With the Build.xml template, the installation can also point to the hashlink sources, provided that they contain binaries generated by `hl.sln`.


OKEAMAH (approved):

kLabz:
Small documentation / default value handling issue:
> Set hlgen.makefile for automatic native compilation

But adding `-D hlgen.makefile` (without setting a value) gives:
> Automatic native compilation not yet implemented for 1

If picking a default value is not something we want, it would be nice to list correct values in this error message I think?

tobil4sk:
> If picking a default value is not something we want, it would be nice to list correct values in this error message I think?

The default value was working before, I just forgot about it with my changes so this should be fixed. Maybe it should be updated though:
- vs2019 for windows
- make for other systems

Aside from this, the makefile here doesn't support mac at the moment (mac linker doesn't support `-l:fmt.hdll` syntax). This will have to be fixed before this can be merged. `-lfmt.hdll` doesn't work because hdll libraries don't follow the standard naming of `lib[name].so`. We can instead link with the full path `/usr/local/lib/fmt.hdll`, however, on linux this hardcodes the full path for loading the library, unless `fmt.hdll` is built with `-soname` (already the case with the hashlink cmake file, but not the hashlink make file). So there are two options:

- Use the full path universally, requiring hdlls (including user-made ones) to be built with `-soname` on linux.
- Generate separate linking flags for mac and linux.

Simn:
Yuxiao made me aware of this PR and I'd like to see this made working. Native compilation is the main difficulty I have with HL/C over hxcpp, where everything mostly "just works". I'm not sure what decisions have to be made here, and who has to make them, so I'll just ping @ncannasse and @yuxiaomao and hope for the best!

ncannasse:
At the moment genhl outputs a file that gives all the needed information
then call the haxelib hashlink module.  I guess the implementation for auto
compilation would go there

Le mer. 9 avr. 2025, 11:59, Simon Krajewski ***@***.***> a
écrit :

yuxiaomao:
This PR is ready for me now, with
- Automatic compilation with make/hxcpp/vs2019/2022 templates
- Mac uses separate linking flags with absolute path to hdll (tested on a Mac with Rosetta)
- Suggests default hlgen.makefile value vs2019 (win) / make (others)

If I'm motivated I can eventually add some CI for this with a link to fmt.hdll (e.g. `haxe.crypto.Sha1.encode("Hello world")`)

PR #709, opened 2024-09-06
[cmake] Standardise install tree for windows build

tobil4sk:
As discussed in https://github.com/HaxeFoundation/hashlink/pull/705#issuecomment-2333771386, this changes the install tree for windows builds in cmake to match the vs2019 generated packages, rather than in separate `bin` and `lib` directories.

This way a cmake install of hashlink will work with the visual studio hlc templates.

I've also changed the default install location on windows to be `C:\HaxeToolkit\hashlink`.


yuxiaomao:
Thank you! Before merge, I'd like to know, (as your PR only affect Windows build,) if it's better to also change the path for Linux/OSX build - or not?

tobil4sk:
> Thank you! Before merge, I'd like to know, (as your PR only affect Windows build,) if it's better to also change the path for Linux/OSX build - or not?

On Mac and Linux it's not possible to stop installing to `bin` and `lib`, this is required for the binaries to run and load correctly.
It would be possible to override this just for packaging (rather than install), but right now the cmake packages are not released officially anywhere (the nightly build only releases the make builds) and this would not be useful for the same reasons as on Windows.

PR #712, opened 2024-09-10
[make] Add "hashlink" library to release script

tobil4sk:


yuxiaomao:
I think hashlink lib will also need `/hlmem`, it is hide's dependency (for memory dump viewer).

PR #713, opened 2024-09-20
Ssl: makefile add static mbedtls build

yuxiaomao:
Allow building ssl.hdll with embedded mbedtls source from hashlink instead of system lib, with `make SSL_STATIC=1`.

This code is useful when the build machine does not have a recent version of mbedtls, and would like to use the one in hashlink - which might be more recent and more secure (e.g. for game distribution reason).

Tested on Linux and seems to work. Don't know how the mbedtls license works here, but it's already the case when building on Windows.


Makefile
```diff
@@ -310,6 +316,9 @@ codesign_osx:
 
 clean_o:
 	rm -f ${STD} ${BOOT} ${RUNTIME} ${PCRE} ${HL} ${FMT} ${SDL} ${SSL} ${OPENAL} ${UI} ${UV} ${MYSQL} ${SQLITE} ${HL_DEBUG}
+ifdef SSL_STATIC
+	make -C "libs/ssl" clean
```
tobil4sk, line 320: Would be best to use `${MAKE}` here, as used above.

libs/ssl/Makefile
```diff
@@ -0,0 +1,132 @@
+INCLUDES_BASE_DIR = ../../include
+CFLAGS = -Wall -O3 -std=c11 -m64 -fPIC -pthread -fno-omit-frame-pointer
+CFLAGS += -fvisibility=hidden -I. -I${INCLUDES_BASE_DIR}/mbedtls/include -I ../../src
+CFLAGS += -D MBEDTLS_USER_CONFIG_FILE=\"mbedtls_user_config.h\"
+LFLAGS = -L../.. -lhl
+
+UNAME := $(shell uname)
+ifeq ($(UNAME),Darwin)
+CFLAGS += -framework Security -framework CoreFoundation
```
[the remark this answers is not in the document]
yuxiaomao: Yes you're right! I'm not familiar with osx at all :x

yuxiaomao:
Thank you for the review and suggestions^^ I also think it's proper to use the main Makefile (and avoid copying cflags/lflags around). I can probably also reduce mbedtls `.o` occupying all that lines similar to PCRE/STD (previously I was just replacing all from the cmakelists)

Makefile
```diff
@@ -221,8 +267,16 @@ sdl: ${SDL} libhl
 openal: ${OPENAL} libhl
 	${CC} ${CFLAGS} -shared -o openal.hdll ${OPENAL} ${LIBFLAGS} -L. -lhl $(LIBOPENAL)
 
+./include/mbedtls/%.o: ./include/mbedtls/%.c
+	${CC} ${CFLAGS} -o $@ -c $< ${SSL_CFLAGS}
+
+# force rebuild ssl.o in case we mix SSL_STATIC with normal build
+.PHONY: libs/ssl/ssl.o
+libs/ssl/ssl.o: libs/ssl/ssl.c
+	${CC} ${CFLAGS} -o $@ -c $< ${SSL_CFLAGS}
```
tobil4sk, line 276: Would it work if the `libs/ssl/ssl.c` dependency is removed? like this:
```makefile
libs/ssl/ssl.o:
	${CC} ${CFLAGS} -o $@ -c libs/ssl/ssl.c ${SSL_CFLAGS}
```

.PHONY implies that `libs/ssl/ssl.o` doesn't exist in the file system: https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

yuxiaomao: The rule `libs/ssl/ssl.o` will not be executed if a file of the same name already exists in the filesystem, and the prerequisite has not changed after the modified time of the `.o`.
.PHONY force it to ignore the file.

yuxiaomao: So removing the dependency will not force the rebuild, as ssl.o already exists;
it's possible to put `rm -f libs/ssl/ssl.o` somewhere to force the rebuild, but I didn't find a good place yet

tobil4sk: > The rule libs/ssl/ssl.o will not be executed if a file of the same name already exists in the filesystem, and all the prerequisite's modification time is before the modification time of the .o.

Makes sense! I guess `.PHONY` works for now then. A better solution might be possible if we write the value of `SSL_STATIC` to a file whenever it changes, and then add that file as a target dependency.

yuxiaomao: Hum add a tmp file can be a good choice, I've already seen such usage in other projects, let me check

yuxiaomao: I think use .PHONY is simpler in this case, even it's not very flexible (e.g. replace `ssl.o` to `%.o` will not work, add other c files will probably not work either). Let us stay with it for now, and we can improve it later

tobil4sk: Sounds good!

PR #714, opened 2024-09-24
Switch to cmake artifacts for release builds

sys9kdr:
Currently, make builds are not working(e.g. missing lib folders, bin folders...). So switch from make artifacts to cmake one.


sys9kdr:
ref #715

sys9kdr:
this pr make no sense. ref: https://github.com/HaxeFoundation/hashlink/issues/715#issuecomment-2378908754

PR #717, opened 2024-10-11
Fix compile issue for iOS

rcstuber:
"hl_thread_set_name" also isn't available on iOS. I would assume tvOS too, but could not test myself


yuxiaomao:
I have fixed the CI and it should be good now

PR #721, opened 2024-10-25
[hlc] Use precompiled headers in vs templates

tobil4sk:
Instead of having to recompile the headers for each individual source file, the headers are now compiled once for the whole project. This gives a further speed up from ~7 seconds to ~3 seconds.

### DIFF
--- a/other/haxelib/templates/vs2015/__file__.vcxproj
+++ b/other/haxelib/templates/vs2015/__file__.vcxproj
@@ -99,6 +99,8 @@
       <SDLCheck>true</SDLCheck>
       <ObjectFileName>$(IntDir)\%(RelativeDir)</ObjectFileName>
       <MultiProcessorCompilation>true</MultiProcessorCompilation>
+      <PrecompiledHeader>Use</PrecompiledHeader>
+      <ForcedIncludeFiles>stdafx.h</ForcedIncludeFiles>
     </ClCompile>
     <Link>
       <SubSystem>Console</SubSystem>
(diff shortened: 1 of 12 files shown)


ncannasse:
Last time I tried that on our projects it turned out to be a big slow down
instead. But @yuxiaomao can confirm that

yuxiaomao:
Tested on shiro's project on my machine. It's better than the current template, and can be useful if we merge; but is still a "big slow down" compared to the 1 file strategy that is currently in use for the project.

| Template | Time |
| -------- | ------- |
| Current vs2019 template with your /Fo optimization | 828-862s (14 mins) |
| With this PR | 330-372s (6 mins) |
| Current ci template (all-in-one C file, no HL_MAKE) | 57-91s (1.5 mins) |

(Last week I tested the current template and it takes me 408s (7mins), still worse than this PR but take less time. I suppose the perf is never stable :/)

Simn:
Comparing a single-file compilation to this doesn't seem particularly fair. If you do that you'd have to take into the account incremental compilation, which should make a huge difference.

yuxiaomao:
Agree. My point is this PR is good and we can merge it.
And the comparison for the single-file is only for comparison, as @tobil4sk discussed with me previously about having a template that can choose to activate single-file compilation with some `-D hlgen` flag.

ncannasse:
Yes but we found actually difficult to keep having exactly same output when
going through compiler cache. Maybe it's better now.

Le lun. 28 oct. 2024 à 10:02, Simon Krajewski ***@***.***> a
écrit :

PR #727, opened 2024-11-25
hlmem: Add type id to print and allow doing locate on type ids

Speedphoenix:
This allows doing locates on types containing spaces

With -c, this displays like this 
image

And then doing `locate #45319` yields things referencing the `Function(Void -> Void)` type

### DIFF
--- a/other/haxelib/hlmem/Memory.hx
+++ b/other/haxelib/hlmem/Memory.hx
@@ -53,8 +53,10 @@ class Stats {
 	}
 
 	public static function getTypeString(mem : Memory, id : Int){
-		var t = mem.types[id & 0xFFFFFF];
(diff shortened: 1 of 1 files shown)


ncannasse:
I think functions should be identified by their implementation instead of their type, this is in discussion with @yuxiaomao

Speedphoenix:
This is actually less about functions and more about any time that has spaces or that is just very long

For example
image
image

Using heaps' hxsl shaders yields some of these
image
image

There are also some structs that fit on 8 lines or more in fullscreen 1440p like this hxbit object (not the only one) 
image

and also some long types without spaces that are fastidious to type or paste without the id
image


Being able to locate functions before their implementation is available is convenient, but more of an added perk than the initial intention

PR #733, opened 2024-12-25
[make] Clean up homebrew package config

tobil4sk:
We do not need to add all the individual include/library paths, most of them are linked to the main homebrew prefix so we can just add that instead.

Only openal-soft is keg-only so it doesn't get linked to the main prefix.

Previous version also ended up with a lot of duplicate include paths being added.


yuxiaomao:
I'm not familiar with homebrew but I'll merge as it does not break any build x)

PR #737, opened 2024-12-30
[hlc] Add jumbo build for vs template

yuxiaomao:
Add jumbo build (aka unity build but I prefer avoid using "unity") for hlc templates and use it by default for better hlc compilation speed with Visual Studio (especially when use hl/jit for dev and hl/c for release).
If incremental build is preferred, jumbo build can be disabled by setting `-D hlgen.nojumbo`.

@tobil4sk made a couple of tests and it turns out that this all-in-one build strategy is still the quickest and is a known approach. It's already used internally in Shiro's project but was made by manually editing vcsproj file.

### DIFF
--- a/other/haxelib/Run.hx
+++ b/other/haxelib/Run.hx
@@ -46,7 +46,7 @@ class Build {
 			tpl = "vs2015";
 		var srcDir = tpl;
 		var targetDir = config.defines.get("hlgen.makefilepath");
-		var jumboBuild = config.defines.get("hlgen.nojumbo") == null;
+		var jumboBuild = config.defines.get("hlgen.makefile.jumbo");
 		var relDir = "";
 		if( targetDir == null )
 			targetDir = this.targetDir;
@@ -126,7 +126,7 @@ class Build {
 				}
 				var content = sys.io.File.getContent(srcPath);
 				var tpl = new haxe.Template(content);
-				content = tpl.execute({
+				var context = {
 					name : this.name,
 					libraries : [for( l in config.libs ) if( l != "std" ) { name : l }],
 					files : files,
@@ -135,7 +135,8 @@ class Build {
 					cfiles : [for( f in files ) if( StringTools.endsWith(f.path,".c") ) f],
 					hfiles : [for( f in files ) if( StringTools.endsWith(f.path,".h") ) f],
 					jumboBuild : jumboBuild,
-				},{
+				};
+				var macros = {
 					makeUID : function(_,s:String) {
 						var sha1 = haxe.crypto.Sha1.encode(s);
 						sha1 = sha1.toUpperCase();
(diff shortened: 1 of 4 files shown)


tobil4sk:
Should we maybe have a single define for enabling or disabling them? Something like this, but I'm not sure if there is a standard for using `yes`/`no` in haxe defines yet:
```hxml
-D hlgen.jumbo=yes
-D hlgen.jumbo=no
```
This will make it easier to have a different default value for this define based on which template we are using, as the trade off will be different for other compilers.

Currently, these template related defines all have makefile in the name, should we continue with this standard? `-D hlgen.makefile.jumbo`?

It may also make sense to rename the `HL_MAKE` define accordingly to something like `HL_JUMBO_BUILD` which makes more sense with the new behaviour (using an "hlgen.makefile" template no longer implies `HL_MAKE` will be set). This would require changing the generated hlc and both would have to be supported for now.

yuxiaomao:
I don't think `=yes` / `=no` is standard, but we can use something like that (or enable/disable). However I'm not sure how we're supposed to configure this default value for a given template: perhaps by sending both "forceJumbo" "forceNoJumbo" variables?

> This would require changing the generated hlc and both would have to be supported for now.

I'm not sure that I understand you here. `HL_MAKE` is actually `HL_NO_JUMBO` and is only used by VS template for now. For your PR adding a makefile incremental build template, it can just define `HL_NO_JUMBO` too.

yuxiaomao:
Changed to `-D hlgen.makefile.jumbo=true` with `setDefaultJumboBuild(true)` (name can be adjust). Comparing bool in template context after set doesn't work well for me, so it's currently just a string.

I agree that `HL_MAKE` does not describe what it does now, but it would cause compatibility problem and would require update haxe too.

tobil4sk:
> Changed to -D hlgen.makefile.jumbo=true with setDefaultJumboBuild(true) (name can be adjust). Comparing bool in template context after set doesn't work well for me, so it's currently just a string.

This looks good!

> I agree that HL_MAKE does not describe what it does now, but it would cause compatibility problem and would require update haxe too.

Perhaps in the future we can sort this out in a backwards compatible way by checking the version in hlc.json.

yuxiaomao:
As jumbo is a string now, would it make more sens to have `-D hlgen.makefile.mode=jumbo` and `$$setDefaultMode(jumbo)`? It can have other values such as nojumbo (or incremental). This leaves the possibility to support other "modes" in a templates - I can't see a real use case for msvc templates now and it's not easy to handle, but for example "flat"

yuxiaomao:
Ok. I'm not very satisfied with the comparison using `== "true"`, but I didn't find a better solution. Let's say this PR is ok for me now.
Will merge next week if no one is against.

Issue #730, opened 2024-12-07
Hashlink floating point exception with remainder operator


PXshadow:
```haxe
function main() {
	var x = (-1 << 31);
	var y = -1;
	trace(x % y);
}
```
Haxe nightly: c4868e5
Hashlink: 1.14.0

Apprentice-Alchemist:
In this case floating point exception actually means integer overflow.
This is an edge case that needs to be worked around either in genhl or in the JIT.

yuxiaomao:
I confirm that it's an integer overflow, and can be reproduce with hl/c too if the optimization is disabled.
This is indeed an edge case :(

> https://wiki.sei.cmu.edu/confluence/display/c/INT33-C.+Ensure+that+division+and+remainder+operations+do+not+result+in+divide-by-zero+errors
> Overflow can occur during a remainder operation when the dividend is equal to the minimum (negative) value for the signed integer type and the divisor is equal to −1.

ncannasse:
Interesting, that's a new class of exception that I just learnt.
In 32 bits it only occurs with these two values exactly, so I guess the best choice is to generate HL/C + JIT code to test divide by (-1) after we checked 0 and if it's the case perform a IMULT instead of a IDIV.

yuxiaomao:
I can't see a case where `% -1` does not return 0, except when overflow occurs; so I'm going to say : `r = (b == 0 || b == -1) ? 0 : a % b` in HLC(haxe)/JIT(hashlink)

yuxiaomao:
Nicolas explained to me x)

OSMod: `r = (b == 0 || b == -1) ? 0 : a % b`
OSDiv:  `r = (b == 0 || b == -1) ? a * b : a / b` with IMUL

OSDiv can be triggered with the following code:
```haxe
static function main() {
	var x = -1 << 31;
	var y = -1;
	var z = Std.int(x / y);
	trace(z);
}
```

PR #740, opened 2025-01-09
[jit] Fix OSMOD/OSDIV overflow with INT_MIN / -1

yuxiaomao:
Related Issue: https://github.com/HaxeFoundation/hashlink/issues/730

I don't really understand why `load(ctx, out, a)` doesn't work well when I added it before the "normal" case, but works in the current version. Otherwise haxe unit tests (and some added in my haxe PR) seems working now.

Some review would be helpful @ncannasse


ncannasse:
Seems good to me

Issue #735, opened 2024-12-26
Segfault/Access Violation when calling local function made with `Reflect.makeVarArgs` in another function


NotHyper-474:
Haxe 4.3.6
HashLink 1.14.0
Repro code: https://try.haxe.org/#F21Ca171
If I explicitly mark `func` as `Dynamic` it works without issues.
Other targets work as expected.

I also noticed that on Windows it'll throw an error instead.
```
source/Main.hx:7: Haxe is great!
Uncaught exception: Access violation
Called from $Main.testFunc(Main.hx:17)
Called from $Main.main(Main.hx:13)
Called from .init(?:1)
```

yuxiaomao:
I confirm that I can repro with your example with latest hashlink.

I note the example here just in case:

```haxe
class Main {
	static function main() {
		var func = function(args:Array<Dynamic>):Void {
			trace("Haxe is great!");
		};

		var f2 /*: Dynamic*/ = Reflect.makeVarArgs(func);
		f2([]); // Works as expected
		testFunc(f2);
	}

	static function testFunc(func:Dynamic) {
		func([]); // Segfaults if f2 is not type explicitly as Dynamic
	}
}
```

NotHyper-474:
Interestingly your example also only fails if `f2([]);` is present.

yuxiaomao:
It also only fails if `:Dynamic` is present in `testFunc(func:Dynamic)`.
Seems to be caused by f2's type, it's typed as `Array<Dyn>->Void` (caused by `f2([])`) and passed as `Dyn`, and hashlink somehow don't like it. I'll see if I can find some fix.

jacobalbano:
I encountered a variation of this issue as well, with no need to use reflection or varargs.
This is as minimal as I've been able to make it while maintaining the crash.

## Main.hx
```haxe

enum TestState {
	Foo;
}

class Main {
	private var statemachine:StateMachine<TestState>;
	private var value:Bool;

	public function new() {
		statemachine = new StateMachine();
		statemachine.addState(Foo, { onUpdate: makeFunction(true) });
		statemachine.changeState(Foo);
	}
	
	function makeFunction(value:Bool) {
		return () -> {
			trace('new value is $value');
			this.value = value;
			// trace('if this trace runs, no error');
		};
	}

	static function main() {
		var m = new Main();
		m.statemachine.update();
	}
  }

typedef StateHandler<TResult> = {
    public function onUpdate():TResult;
}

[...]

PR #755, opened 2025-02-25
Fix makeVarArgs cast to func and call as dyn segfault

yuxiaomao:
Fixes https://github.com/HaxeFoundation/hashlink/issues/735

I'm not sure if this is the best solution, but I think at least I did not break it more than before.

The problem in the original issue is that a `makeVarArgs` function is cast to another HFUN type before calling it as `Dynamic` (hl_dyn_call). And the other HFUN type does not have a parent, caused access violation.
This will just prevent null parent crash, and then remove wrapper before calling makeVarArgs.

### DIFF
--- a/src/std/fun.c
+++ b/src/std/fun.c
@@ -149,6 +149,9 @@ HL_PRIM vdynamic* hl_call_method( vdynamic *c, varray *args ) {
 	if( args->size > HL_MAX_ARGS )
 		hl_error("Too many arguments");
 	if( cl->hasValue ) {
+		while( cl->hasValue == 2 ) {
+			cl = ((vclosure_wrapper*)cl)->wrappedFun;
+		}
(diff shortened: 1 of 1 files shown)


yuxiaomao:
It seems that deal in callMethod is useful, otherwise Reflect.callMethod will error in this example

```
class Test {
	static function main() {
		var func = function(args:Array<Dynamic>):Void {
			trace("Haxe is great!" + args);
		};

		var f2 /*: Dynamic*/ = Reflect.makeVarArgs(func);
		f2("a"); // Works as expected
		testFunc(f2);
		Reflect.callMethod(null, f2, ["d"]); // Error if callMethod does not deal with hasValue 2
	}

	static function testFunc(func : Dynamic) {
		func("b", "c"); // Segfaults if f2 is not type explicitly as Dynamic
	}
}
```

Issue #747, opened 2025-01-28
GC crash


ncannasse:
I could accidentally reproduce a GC crash, where a block `0x0000027200000000` was accessed in gc_flush_mark (p = *block++) triggering an access violation.

GC_GET_PAGE(block) was NULL but gc_hash was discarding higher bits.
To be investigated

PR #756, opened 2025-02-27
[gc] add some check for gc_flush_mark

yuxiaomao:
Maybe for https://github.com/HaxeFoundation/hashlink/issues/747

When block is not actually a pointer.

Ping @ncannasse just in case

### DIFF
--- a/src/allocator.c
+++ b/src/allocator.c
@@ -585,7 +585,7 @@ static int gc_allocator_get_block_interior( gc_pheader *page, void **block ) {
 			bid--;
 		}
 		int size = page->alloc.sizes[bid];
-		if( size <= (start - bid) )
+		if( (start - bid) >= size )
 			return -1;
 	}
 	*block = page->base + bid * page->alloc.block_size;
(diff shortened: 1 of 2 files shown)


yuxiaomao:
There are 3 conditions now,
1. `if( !page || !INPAGE(block,page) )`. Inspired by the fact that most `GC_GET_PAGE` is followed with such test except for `block`.
2. `if( !t || t->kind < 0 || t->kind >= HLAST )`. Inspired by the `GC_DEBUG` code just above checking `t->kind` value in Dynamic page.
3. `if( !INPAGE(block,page) )` in `while( pos < nwords )`. Inspired by the pointer address end in 000 in #747 .

We have someone who can reproduce an access violation sometimes, it's likely a GC bug too. Here is some observations until now:
- 1 => still have access violation
- 1 + 2 + 3 => error in 2, e.g. with `size=24`, `block=0x0000022F4D328B98`
- 1 + 3 => ? (testing)

yuxiaomao:
For now, we have 1 condition that adresse the access violation
- `if( bid >= page->alloc.max_blocks )`: invalid bid

2 conditions that reduce false positive
- `if( bid < page->alloc.first_block )`: invalid bid
- `if( (start - bid) >= size )` in `gc_allocator_get_block_interior`: invalid bid found when searching backward

1 condition for optimization
- `if( !t )`: corresponding block is never used

PR #760, opened 2025-03-11
Update README.md

B-Interactive:
Added Arch Linux make and install instructions.  Dependencies are automatically installed in this process, so it was a little tricky to decide where to place this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,20 @@ To install hashlink binaries on your system you can then call:
 
 `make install`
 
+**On Arch Linux:**
+
+Make and install it using the [Arch User Repository](https://aur.archlinux.org/packages/hashlink) (AUR):
+
+```
+git clone https://aur.archlinux.org/hashlink.git
+cd hashlink
+makepkg -sri
+```
+
+Or using an AUR helper such as [yay](https://aur.archlinux.org/packages/yay-bin):
+
+`yay -S hashlink`
+
 ## Building on Windows
 
 Open `hl.sln` using Visual Studio C++ and compile.


yuxiaomao:
I'm not familiar with Arch Linux, but what you've provide does not seems to be a build from source (git latest) instruction, and can be confusing if placed in README.
If you think it's very important for Arch Linux user, I suggest you to put it into the wiki: https://github.com/HaxeFoundation/hashlink/wiki/Building-and-Installing#how-to-install

B-Interactive:
It would be building from source (pulling from `https://github.com/HaxeFoundation/hashlink/archive/${pkgver}.tar.gz` which would currently represent release 1.14).  

This is the Arch way of things.  One does not build from source independent of the package management system, one leverages the package management system to build from source.  The [PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=hashlink) file for this package shows where the source is pulled from, and how it is built (make) and installed, as well as any dependencies that will be installed with it.

If the README.md is for how to build against the latest commit, then a simple adjustment to installing [hashlink-git](https://aur.archlinux.org/packages/hashlink-git) (latest commit) instead of [hashlink](https://aur.archlinux.org/packages/hashlink) (latest release) is all that is needed.

I don't mean to wrestle this change in though.  I simply offered it for the sake of thoroughness, and as the package maintainer for [hashlink-git](https://aur.archlinux.org/packages/hashlink-git) and fond user of Haxe.

tobil4sk:
> One does not build from source independent of the package management system, one leverages the package management system to build from source.

The instructions in the README are intended for people who have downloaded the sources for the purpose of building them directly, for example if you're working on changes to hashlink.

That being said, platform specific instructions like this are useful and should have a place, though I would agree the wiki should be that place (otherwise it gets unwieldy). It could be linked from the README to be more discoverable for users who are just looking to install hashlink on their system rather than building from source.

PR #778, opened 2025-04-20
[gc] remove `memmove` usage in hl_remove_roots

mrgaturus:
because the insertion order of `gc_roots` list doesn't matter it's possible remove the `memmove` usage and just move the last root from the list to the found root position.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ x64
 *.a
 *.opendb
 *.VC.db
+.cache
 .vs
 /hl
 /include/sdl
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -21,6 +21,7 @@ x64
 *.a
 *.opendb
 *.VC.db
+.cache
```
yuxiaomao, line 24: Could you explain why this gitignore entry is needed ?

yuxiaomao:
Thank you!

PR #779, opened 2025-04-22
Add initial Emscripten compilation support.

tritao:
Just gets the basics of Hashlink to compile and run on Emscripten platform.


yuxiaomao:
I did not test but seems legit and can't harm x)

PR #782, opened 2025-04-25
Check for __customStack on haxe.Exception

Simn:
This allows printing custom exception stacks on uncaught exceptions in case we have an instance of `haxe.Exception` which has `_hx_customStack` set. This will be useful for the coroutine implementation.

It's currently broken because `uprintf` should apparently not be used like that (it works for me locally though), but I'm sure Yuxiao can save me!


yuxiaomao:
Note
- Related : https://github.com/Aidan63/haxe/pull/50#issuecomment-2830075393
- Need to check hlc
- Need to double check the name `_hx_customStack` or `__customStack` etc

PR #784, opened 2025-04-30
Fix compile error

Apprentice-Alchemist:
`fun_field_name` returns `const uchar *` (`const char16_t`) not `vbyte *` (`unsigned char *`)

GCC 14 has made the `incompatible-pointer-types` warning a hard error.


yuxiaomao:
Thank you!

PR #789, opened 2025-05-19
[jit] Make sure `ctx->currentPos > 0` before any register allocation …

### LINKED ISSUE #787: "JIT ERROR 0 (jit.c line 968)" Out of registers !  (Occurs all >1.15.0)
Hello,

I am encountering the error in the issue title when trying to run my HL app.

"JIT ERROR 0 (jit.c line 968)" 

That line has a comment: Out of registers ! 

The app is here:
https://github.com/inert-aesthetic/chase/

The error happens when this class is included:

https://github.com/inert-aesthetic/chase/blob/master/src/game/systems/BoundsSystem.hx

I have not had success with a minimal repro, because just copying the content of the build macro's output manually into the class will avoid the error.

(comment metadata in line 6, uncomment lines 10-22 and 81-end to see no error)

If this is hard to track, I can provide dump, compiled .hl (working and unworking), etc. Please ask!

I have found that the error does not change anywhere from git latest hl to 1.15.0. 
Any earlier and the API mismatches Heaps so I haven't checked.

Apprentice-Alchemist:
…can happen.

`copy` can trigger register allocation eg when booleans are involved, so `currentPos` needs to be `>0` before then otherwise `alloc_reg` thinks all registers are locked and fails.

Closes #787

### DIFF
--- a/src/jit.c
+++ b/src/jit.c
@@ -2935,6 +2935,9 @@ int hl_jit_function( jit_ctx *ctx, hl_module *m, hl_function *f ) {
 	ctx->totalRegsSize = size;
 	jit_buf(ctx);
 	ctx->functionPos = BUF_POS();
+	// make sure currentPos is > 0 before any reg allocations happen
+	// otherwise `alloc_reg` thinks that all registers are locked
+	ctx->currentPos = 1;
 	op_enter(ctx);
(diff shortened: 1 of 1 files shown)


yuxiaomao:
Thank you!

Issue #636, opened 2023-12-03
Linking to libopenal.a causes duplicated symbol error


fuzhouch:
Hi,

    I'm working on a project to build Hashlink against 100% static libraries, which targets creating a binary that can be copied and use in any Linux distro. I've done many modules (ui,uv, sdl2, etc) but stuck on openal. When I link against a static libopenal.a, gcc complains a lot of "duplicated symbol" errors.

I have spotted the root cause: it's because AL_IMPORT() and ALC_IMPORT() macro use a trick when loading extensions, that it defines a global function pointer, with same name of corresponding openal-soft functions. Then the function pointers are loaded via alc_load_extensions(). When linking against a static libopenal.a, the function pointer variables and functions have naming conflict, thus causes compiler complains.

I have a [private fix](https://github.com/HaxeFoundation/hashlink/compare/master...fuzhouch:hashlink:fix-openal-staticlink
) by modifying AL_IMPORT() and ALC_IMPORT() macros, to use add function pointers with _dyn suffix.  It verified it with my own [build environment project](https://github.com/fuzhouch/hashbld/). 

However, I'm wondering whether this is an intended coding pattern with existing reason, what we use alGetProcAddress() to load extension functions. Actually, I see many functions use CHECK_EXT() macros, that if extension function does not exist, it just breaks. Is there any scenario that needs to link with libopenal.so?

tobil4sk:
> I have a [private fix](https://github.com/HaxeFoundation/hashlink/compare/master...fuzhouch:hashlink:fix-openal-staticlink) by modifying AL_IMPORT() and ALC_IMPORT() macros, to use add function pointers with _dyn suffix.

A bit late, but I think another way of solving this would have been to mark the definitions in `openal.c` as static. E.g. `#define ALC_IMPORT(fun, t) static t fun`

> What's more, I can see the extension functions have existed in openal-soft and OpenAL, so looks like there's no need to do a if-exist-and-load trick. Or, if we can call the functions directly, it will be even easier.

Here is some information about the purpose of `alGetProcAddress`:

https://github.com/kcat/openal-soft/issues/745

Since merging #800 and #801, it is now possible to link openal statically with cmake using `-DWITH_STATIC_OPENAL=ON`.

The way I did this is by removing the dynamic loading for static builds and linking directly against the functions. According to the openal-soft maintainer on IRC, `alGetProcAddress` would still work fine on static builds, but there is not much point since it is linked statically so it is known at build time which extensions are available.

PR #798, opened 2025-07-27
[cmake] Improve support for android build

### LINKED ISSUE #109: android vs third-party libraries
I had to make some changes to turbojpeg/vorbis/libpng to make fmt compile, because the sources included in the repo are stripped and doesn't include support for ARM architectures and specific optimizations (e.g. NEON), which raises a question:

What's the reccomended way to work with these libs? Judging from CMakeFiles, the included third-party libraries are only used on windows, so I guess on android/ios one must use official distros?

tobil4sk:
- Add [BUILD_SHARED_LIBS](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html) option, allowing static builds of libraries.
- CMakeLists.txt now downloads all required dependencies for android, making the build more convenient. Closes #109.
  - Note SDL2 and openal are linked dynamically, while the others are linked statically. This is because SDL2 has its own JNI_OnLoad which would conflict with hashlink's sys_android.c one, and because openal fails when linking statically: #636.
- Fix various compatibility issues with android
- Add android build to ci to track compatibility (ANDROID_PLATFORM=24 required by libuv)


yuxiaomao:
I can't really review the android part but seems good. Thank you!

tobil4sk:
Thank you!

Issue #716, opened 2024-10-08
Compile Problem HL/C


barisyild:
Haxe Code:
https://github.com/boorik/smartfox-haxe-client/blob/d7a10b57dc389bef19eab7508480cbd9df4739d4/com/smartfoxserver/v2/SmartFox.hx#L939

https://github.com/boorik/smartfox-haxe-client/blob/d7a10b57dc389bef19eab7508480cbd9df4739d4/com/smartfoxserver/v2/core/SFSEvent.hx#L54

C Output:
```c
r13 = (com__smartfoxserver__v2__core__$SFSEvent)g$da560e3;
r7 = r13->HANDSHAKE;
```
Error:
```ex
Export/hl/obj\com/smartfoxserver/v2/SmartFox.c(351): error C2037: left of 'HANDSHAKE' specifies undefined struct/union '_com__smartfoxserver__v2__core__$SFSEvent'
```

yuxiaomao:
Could you provide a minimal example?

barisyild:
> Could you provide a minimal example?

The library link is already above.

tobil4sk:
@barisyild Does this still reproduce for you? I have tried with haxe 4.3.6, 4.3.7 and 5.0.0-preview.1 and couldn't reproduce this error. For me, the `HANDSHAKE` string is correctly inlined:
```c
r2 = (String)s$handshake;
```

barisyild:
> [@barisyild](https://github.com/barisyild) Does this still reproduce for you? I have tried with haxe 4.3.6, 4.3.7 and 5.0.0-preview.1 and couldn't reproduce this error. For me, the `HANDSHAKE` string is correctly inlined:
> 
> r2 = (String)s$handshake;

What is your os, i tried with windows.

tobil4sk:
I'm also testing this on windows 11

barisyild:
Also i can't reproduce, probably fixed.

tobil4sk:
@barisyild Actually, the reason I wasn't able to reproduce was due to `-dce full`. If I remove that, I am able to reproduce it:

```haxe
class Main {
	public static inline var HANDSHAKE:String = "HANDSHAKE";
	public static inline var SOCKET_ERROR:String = "SOCKET_ERROR";

	static function main() {
		trace(Main.HANDSHAKE);
		trace(Main.SOCKET_ERROR);
	}
}
```

```hxml
--dce no
-cp src
-D analyzer-optimize
-main Main
[...]

PR #804, opened 2025-08-02
Remove windows.h include from hl.h

tobil4sk:
By removing the direct call to `IsDebuggerPresent` in `hl.h` (we can call `hl_detect_debugger` instead), we can avoid the `windows.h` include in `hl.h`.

This avoids issues related to `windows.h` when including `hl.h`, such as #716. It also allows removing some `_WINSOCKAPI_` hacks.

`windows.h` must still be included in `hlc_main.c` (otherwise `DbgHelp.h` is broken), which can still impact user code in jumbo builds (see #737). However, we can also minimise the impact there by applying `WIN32_LEAN_AND_MEAN`.

Defining `_GUID` breaks windows headers, so I've had to undefine it in a bunch of additional places (see also: 65e1bc4cfe143958d871522d5d64f9eecf546025).


ncannasse:
I don't like much the resulting code. The original issue is a C language issue and even if we don't include windows.h there will be some identifiers that are valid in Haxe but don't compile in C.

Maybe a better solution would be a define in Haxe that escapes such identifiers in C output code for example -D hlc-escape=HANDSHAKE;HL_WIN 

I'm still in favor on adding most common identifiers to be undef by default as we do now. This would be for more rare keywords.

tobil4sk:
> I don't like much the resulting code.

Which part of the resulting code do you not like? Can none of it be improved?

> The original issue is a C language issue and even if we don't include windows.h there will be some identifiers that are valid in Haxe but don't compile in C.

This is true, however, `hl.h` is a public header that is included by hlc programs and user hdlls, not just by the hashlink runtime. Including the massive `windows.h` in a public header like this creates many problems, such as unwanted macros, needing to define `_WINSOCKAPI_` before including `hl.h` to be able to use `winsock2`, etc.  Other users of `hl.h` may not need `windows.h`, so they should not be forced to have it just because it is required internally in libhl/hljit.

I think public headers should be minimal anyway, to provide most flexibility for users and also avoid the compiler processing unnecessary code.

Removing `windows.h` will reduce the occurrence of these broken hlc identifiers by a very large amount, even if it still leaves some cases unhandled.

> I'm still in favor on adding most common identifiers to be undef by default as we do now. This would be for more rare keywords.

While this will be necessary for builtin compiler defines, this is a much less effective solution than preventing the defines where possible. `#undef` would require a constant game of catch up. In my opinion, having to undefine the entire header after including it is not as elegant as simply avoiding an unneeded include in the first place.

---

Please could you reconsider this? I can move those exact includes as they were from `hl.h` into a private `hl-windows.h` header to avoid the messy duplicate `ifdef`s if that's what the issue is. This way there would be more minimal code changes while still respecting the fact that `hl.h` is a public header, used outside of libhl/hljit. Or there could even be something like `HL_NO_INCLUDE_WINDOWS_H`, which would not be ideal, but I'd still prefer to be able to avoid `hl.h` including `windows.h`, even if it's not by default.

ncannasse:
Well I guess we could have a hlsystem.h that does exactly what hl.h
currently does (wrt windows.h / xdk inclusion ) and that can be included in
other hl source files.
This could be not windows specific though and also allows to include some
linux/osx includes that are not used for hl.h definitions but only for
implementation, if any.

tobil4sk:
> Well I guess we could have a hlsystem.h that does exactly what hl.h
currently does (wrt windows.h / xdk inclusion ) and that can be included in
other hl source files.

This sounds like a good solution to me, please could you reopen the PR so I can push an updated version?

yuxiaomao:
This PR is ok for me now and we can deal with the `#undef _GUID` in another PR, I'll merge it tomorrow if no one is against.

Issue #331, opened 2019-11-24
HL/C compilation fails then using hldx


YRTV:
OS Windows 10 1903 Build 18362,476
Hashlink version 1.10 from release page and latest git master branch tested.
HL/C code generated from following snippet will not compile, bytecode(.hl) version works.

```
import dx.Window;

class Main {

    static function main() {
         new Window("dxwindow", 640, 480);
     }
}
```

MSVC Log
```
**********************************************************************
** Visual Studio 2019 Developer Command Prompt v16.3.10
** Copyright (c) 2019 Microsoft Corporation
**********************************************************************
[vcvarsall.bat] Environment initialized for: 'x64_x86'

C:\Program Files (x86)\Microsoft Visual Studio\2019\Community>cd /d D:\yurii\Documents\HaxeThings\hldxfail\out

D:\yurii\Documents\HaxeThings\hldxfail\out>cl main.c libhl.lib directx.lib /Ox /I . /I %HASHLINK_SRC% /link /LIBPATH:%HASHLINK% /OUT:main_cl.exe
Microsoft (R) C/C++ Optimizing Compiler Version 19.23.28107 for x86
Copyright (C) Microsoft Corporation.  All rights reserved.

main.c
.\dx/Window.h(21): error C2059: syntax error: '<parameter-list>'
.\dx/Window.h(24): error C2059: syntax error: '}'
.\dx/Window.c(294): error C2037: left of 'windows' specifies undefined struct/union '_dx__$Window'
.\dx/Window.c(378): error C2037: left of 'windows' specifies undefined struct/union '_dx__$Window'
.\hl/init.c(557): error C2037: left of 'windows' specifies undefined struct/union '_dx__$Window'
.\hl/init.c(560): error C2059: syntax error: '('
.\hl/init.c(562): error C2275: 'dx__$Window': illegal use of this type as an expression
.\dx/Window.h(4): note: see declaration of 'dx__$Window'
.\hl/init.c(562): error C2146: syntax error: missing ';' before identifier 'g$_dx_Window'
.\hl/init.c(563): error C2037: left of 'HIDDEN' specifies undefined struct/union '_dx__$Window'
.\hl/init.c(566): error C2037: left of 'RESIZABLE' specifies undefined struct/union '_dx__$Window'
```

MinGW Log
```
yurii MINGW32 ~
$ cd D:\\yurii\\Documents\\HaxeThings\\hldxfail\\out

[...]

PR #809, opened 2025-08-06
[hlc] move some windows.h only undefs to hlc_main

yuxiaomao:
Following
- https://github.com/HaxeFoundation/hashlink/pull/804

which moves `#include <windows.h>` from `hlc.h` to `hlc_main.h`, move `undef` that are `<windows.h> + WIN32_LEAN_AND_MEAN` only into `hlc_main.h`

Related previous commit that added undefs:
- https://github.com/HaxeFoundation/hashlink/commit/63763dc3d8fcbae5a02a731217a60ad93eb746a1
- https://github.com/HaxeFoundation/hashlink/commit/2a205c489917b15b58eeb048ef66f58cf2bd0805
- https://github.com/HaxeFoundation/hashlink/pull/753

### DIFF
--- a/src/hlc.h
+++ b/src/hlc.h
@@ -34,14 +34,10 @@
 #ifdef HLC_BOOT
 
 // undefine some commonly used names that can clash with class/var name
-#undef CONST
 #undef stdin
 #undef stdout
 #undef stderr
 #undef EOF
-#undef HIDDEN
-#undef RESIZABLE
-#undef __SIGN
 #undef INT_MAX
 #undef INT_MIN
 #undef BIG_ENDIAN


yuxiaomao:
Note:
- The following undefs are removed because they do not seems necessary for Windows/Linux: `HIDDEN`, `RESIZABLE`, ~~`__SIGN`~~
  - `HIDDEN`, `RESIZABLE` was added in https://github.com/HaxeFoundation/hashlink/issues/331
  - ~~`__SIGN` was added in https://github.com/HaxeFoundation/hashlink/pull/393~~
- The following undefs are needed by Mac: `BIG_ENDIEN`, `LITTLE_ENDIEN`, `TRUE`, `FALSE`, `__SIGN`
  - ref https://github.com/HaxeFoundation/haxe/issues/11419
- The following defines are present with and without windows.h on Windows: `__valid`

PR #813, opened 2025-08-09
[hlc] Fix cli args handling with WINDOWS subsystem

### LINKED ISSUE #454: Calling Sys.args() causes HLC to close when the subsystem is set to "windows" instead of "console"
I changed the subsystem to "WINDOWS" to stop the split second console window from showing up before the main HLC window shows up. The problem is, `Sys.args()` calls now close it immediately.

tobil4sk:
`__wargv` is `NULL` when using `WinMain`. If we want it to be not `NULL`, we need to use `wWinMain`.

Closes #454.

We changed from `__argv` to `__wargv` in d255a90a5296975098bb4132b4f1234de934afa6. This was necessary because the pointer conversion when passing `__argv` is not valid, causing garbled output from `hl_sys_args` and compiler errors/warnings.

The `HL_XBS` version doesn't seem correct to me either, but I don't have the means to test that.

### DIFF
--- a/src/hlc_main.c
+++ b/src/hlc_main.c
@@ -194,7 +194,7 @@ int main(int argc, char *argv[]) {
 }
 
 #if defined(HL_WIN_DESKTOP) && !defined(_CONSOLE)
-int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdLine, int nCmdShow) {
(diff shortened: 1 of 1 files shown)


yuxiaomao:
I think XBS can now use the same call as Windows, I'll commit once verified.
I didn't noticed that the params isn't exactly the same, I should probably fix our internal code too ^^'

yuxiaomao:
XBS commit https://github.com/HaxeFoundation/hashlink/commit/9cbaa4a8a3c8a5acb3dcb218bf1efb7c6d23ec70

PR #821, opened 2025-08-31
[ci] build self-contained macOS binaries

Apprentice-Alchemist:
- add a cmake option to enable dependency downloading (Windows still uses vendored dependencies in all circumstances)
    - OpenAL is dynamically linked by default and the .dylib is included in the install package
    - other dependencies are statically linked
- add FLAT_INSTALL_TREE cmake option to match layout from `make release`
- set rpath on apple platforms
- build macOS binaries on macos-latest with downloaded dependencies and a proper deployment target
    - deployment target for x86-64 is 10.13 (minimum version supported by current macOS SDKs)
    - deployment target for arm64 is 11.0 (Big Sur is the first version to support Apple Silicon)


.github/workflows/build.yml
```diff
@@ -162,16 +168,30 @@ jobs:
...
+              darwinarm64)
+                cmake -B build -G Ninja -DCMAKE_BUILD_TYPE=${{ matrix.cmake_configuration }} \
+                  -DDOWNLOAD_DEPENDENCIES=ON \
+                  -DCMAKE_OSX_ARCHITECTURES=arm64 \
+                  -DCMAKE_OSX_DEPLOYMENT_TARGET=11.0 \
+                  -DFLAT_INSTALL_TREE=ON
+                ;;
               *)
-                cmake . -DCMAKE_BUILD_TYPE=${{ matrix.cmake_configuration }} ;;
+                cmake -B build -DCMAKE_BUILD_TYPE=${{ matrix.cmake_configuration }} ;;
```
yuxiaomao, line 190: Is that useful for Linux to deploy with `FLAT_INSTALL_TREE=ON` too? now that Windows/Mac's cmake builds are flat.

Apprentice-Alchemist: Yes, but I'm leaving that for a future PR.

yuxiaomao:
I could not fully review these cmake changes, but CI is green and Zeta is happy, so seems fine to me x) And thank you!

sebthom:
@Apprentice-Alchemist the **arm64** build seems to be missing the hl binary file.

image

tobil4sk:
@sebthom This is expected, the jit is not supported yet on arm64.

PR #823, opened 2025-09-03
[ci] add gcc14 make on linux

yuxiaomao:
Detect errors such as in
- https://github.com/HaxeFoundation/hashlink/pull/784
- https://github.com/HaxeFoundation/hashlink/pull/822 (<- test without this fix and detect correctly)

Code from
- https://github.com/nim-lang/Nim/blob/b11ceb07018c3521ba94cdb134feefc1abae24d4/.github/workflows/ci_gcc14.yml#L39-L45

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,11 +23,11 @@ jobs:
       matrix:
         target: [linux, darwin, windows]
         architecture: [32, 64, arm64]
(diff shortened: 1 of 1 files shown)


yuxiaomao:
Indeed it works without export CC, as show in my run https://github.com/yuxiaomao/hashlink/actions/runs/17455087109/job/49566955510
- based on master just before #822

PR #831, opened 2025-09-24
Fix mingw build

tobil4sk:
This is an updated version of #482, with some clean up and fixes for new issues that have come up since then.

In particular, `hl_thread_set_name` uses an old technique to change the name of a thread on windows that uses msvc specific syntax: https://learn.microsoft.com/en-us/visualstudio/debugger/tips-for-debugging-threads?view=vs-2022&tabs=csharp#set-a-thread-name-by-throwing-an-exception.

For mingw builds it will use the newer method, which is to use SetThreadDescription: https://learn.microsoft.com/en-us/visualstudio/debugger/tips-for-debugging-threads?view=vs-2022&tabs=csharp#set-a-thread-name-by-using-setthreaddescription. We could also use this on msvc too, but I have left that as is for now.

This allows hl, libhl, and all hdlls except dx12.hdll and video.hdll to compile and link properly.

Closes #482.


N3rdL0rd:
Awesome, I'll merge this into my fork. I'll also work on my own to see if I can get the rest of the hdlls to build correctly, too :)

tobil4sk:
> I'll also work on my own to see if I can get the rest of the hdlls to build correctly, too :)

ui.hdll is just a syntax error, which probably doable, I just got put off by the expression that's causing the error x):
```
.../hashlink/libs/ui/ui_win.c:229:34: error: lvalue required as decrement operand
  229 |                                 *--(int_val*)regs.Rsp = regs.Rip;
      |                                  ^~
```

https://github.com/HaxeFoundation/hashlink/blob/f97d65cb33a19077e6b6ff558eb1a1f6dbca474f/libs/ui/ui_win.c#L229

For most of the other hdlls, it's probably just a question of making sure all the dependencies are installed on your system and adjusting cmake to use the system ones, or to download and build them from source.

N3rdL0rd:
> ui.hdll is just a syntax error, which probably doable, I just got put off by the expression that's causing the error x):

Oh yeah, that's freaky lol

N3rdL0rd:
Here's a few other miscellaneous fixes that managed to get everything working **besides dx12.hdll** for me.

[fix_mingw.patch](https://github.com/user-attachments/files/22519910/fix_mingw.patch)

tobil4sk:
@N3rdL0rd Thanks! I took some of your changes in with some minor tweaks:

- The vendored glext.h is outdated, which is why there is a compiler error. I fixed it by using mingw's system one instead
- For me `hl` links fine without dbghelp, is it required for you?
- On my machine if I set `libhl`'s output name to `libhl`, then it creates `liblibhl.dll`

This leaves video.hdll and dx12.hdll. Video uses a very outdated version of ffmpeg, which is probably the cause, and dx12 requires `dxcompiler` which isn't provided by mingw. All the others are now compiling

N3rdL0rd:
>     * For me `hl` links fine without dbghelp, is it required for you?

Ah whoops, that's left over from my fork, which adds some nicer tracebacks to the signal handler in case of things like segfaults. Ignore that.

>      * On my machine if I set libhl's output name to libhl, then it creates liblibhl.dll

Yeah, I saw the same thing, but changing that fixed an error I experienced earlier on during my experimentation with it. I'll test again and see if it's truly needed now.

> This leaves video.hdll and dx12.hdll. Video uses a very outdated version of ffmpeg, which is probably the cause, and dx12 requires dxcompiler which isn't provided by mingw. All the others are now compiling

I haven't even been able to get video.hdll compiling with MSVC... it might be time to update it, honestly. As for dxcompiler... we might be able to generate a `.a` lib for it with `dlltool` if we can find the location of `dxcompiler.dll` on the fly during CMake configure.

PR #855, opened 2025-12-16
[ci] add dxcompiler.dll, dxil.dll into windows VS build

yuxiaomao:
Currently downloaded hl binaries (Windows + make release) allows user to run with SDL but not with DX, this PR add required DX dlls

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -174,6 +174,12 @@ jobs:
             ;;
         esac
 
+    - name: "Download: DirectX binaries for windows + make"
+      if: matrix.target == 'windows' && matrix.build_system != 'cmake' && matrix.build_system != 'cmake-mingw'
+      run: |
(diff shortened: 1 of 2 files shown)


tobil4sk:
Should we also update the vcxproj files so they contain the `include/dx` library path?

Also, it doesn't have to be handled in this PR, but we could also download the dlls for cmake builds too, and tell cmake to look in include/dx. Not sure if it would work for mingw or not, since the release probably has msvc binaries.

Just about the title, I think it would be less confusing to refer to it as the visual studio build rather than the make build, because make is only used to package the binaries from visual studio.

yuxiaomao:
I think update vcxproj isn't necessary, because the dlls aren't compile time dependences;
It could be useful to download for cmake builds but as there are already other dlls differences, e.g. SDL.dll, I didn't check on cmake

PR #856, opened 2025-12-20
[std] Fix hl_dyni32 for systems with unsigned char

tobil4sk:
On systems where char is unsigned, this function does not behave correctly, as `(char)255` is equal to `-1`, so `hl_dyni32(255)` returns `-1`.


Simn:
TIL that `char` isn't signed by default, unlike `int`.

PR #857, opened 2025-12-20
fix arm64 jit: multiple correctness and safety fixes

JamesKim2998:
## Summary

This PR contains multiple fixes for the ARM64 JIT compiler:

- **V8-V15 callee-saved FP register handling**: Never allocate V8-V15 since the prologue doesn't save them (AAPCS64 compliance)
- **RTMP/RTMP2 (X27/X28) save/restore**: Save callee-saved scratch registers in prologue/epilogue (frame size 80→96 bytes)
- **Switch statement optimization**: Replace O(n) linear scan with O(1) branch table using ADR/BR
- **op_get_mem register clobbering fix**: Use RTMP2 for offset calculation to avoid clobbering base register
- **Exception type filtering**: Properly dereference global pointers for exception type checks
- **hl_jit_free hardening**: Add double-free and use-after-free protection
- **W^X memory protection**: Proper write/execute toggling for Apple Silicon
- **Register spilling safety**: Free scratch registers (X9, V16) before reuse
- **Size encoding fix**: Handle all memory access sizes (1/2/4/8 bytes) correctly
- **OAssert call sequence**: Fix literal pool pattern for indirect calls

## Test plan

- [x] Verified instruction encodings against ARM64 reference manual
- [x] Verified AAPCS64 calling convention compliance
- [ ] Run HashLink test suite on ARM64

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/other/tests/minimal/test_mdbg.c
+++ b/other/tests/minimal/test_mdbg.c
@@ -397,6 +397,127 @@ TEST(verify_cpsr_is_32bit) {
 }
 
 /* ============================================================
+ * Bug #4: ARM64 single-step requires MDSCR_EL1.SS, not CPSR TF
+ *
+ * On ARM64, single-stepping is enabled via MDSCR_EL1.SS (bit 0)
+ * which is accessed via debug registers (REG_DR6 -> __mdscr_el1).
+ *
+ * On x86, single-stepping uses EFLAGS.TF (bit 8).
+ *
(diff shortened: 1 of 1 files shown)


tobil4sk:
Did you mean to open this PR against https://github.com/bmdhacks/hashlink?

Issue #807, opened 2025-08-04
Prefix all defines in hl.h with `HL_`


tobil4sk:
The majority of macros in `hl.h` already follow this rule, however, there are a few that don't, in particular type names for ffi:

https://github.com/HaxeFoundation/hashlink/blob/fbfedd8097d3a2c4869154251f4a160d13a23ac3/src/hl.h#L824-L848

In C, identifiers beginning with an underscore and an uppercase letter [are reserved](https://devblogs.microsoft.com/oldnewthing/20230109-00/?p=107685), so it is not a good idea to be using these. This is why the `#undef` hacks are needed for `_NULL`, `_VOID` and `_STRING`, and why `#undef _GUID` is required in some cases after including `hl.h`: #748 #804.

It would be good to gradually transition to names that aren't reserved and preferably names prefixed with `HL_`. This can be done by providing both defines for now and marking the current ones as deprecated, and adding a flag to turn off the old definitions.

This not only avoids the use of reserved identifiers, but it also makes it easier for the hlc generator to check haxe identifier clashes with `hl.h` macros. It can simply assume that anything starting with `HL_` needs to be mangled, rather than having to be aware of all the types defined by `hl.h`. Related issue: https://github.com/HaxeFoundation/haxe/issues/11419

(As a sidenote, the type name macros are only required for building `hdll`s, and not in `hl` vm code or in `hlc` programs, so it might be worth considering moving these to a separate `hl_ffi.h` header.)

PR #859, opened 2025-12-31
[ffi] add HL_ prefix and move outside of hl.h, allow disable legacy definition

yuxiaomao:
See
- #807 

As suggested added `HL_` prefix and `HL_DISABLE_LEGACY_FFI`


tobil4sk:
This looks good! hl_ffi.h also needs to be included in all packaging/install scripts for this to work.

tobil4sk:
I wonder if it might make sense to do something like this:

hl.h contains legacy ffi definitions only, instead of them being in hl_ffi.h and hl.h including hl_ffi.h. hl_ffi.h only would contain the new definitions, so it is clean. Also hl_ffi.h can define `HL_DISABLE_LEGACY_FFI` before including hl.h. This way if users migrate to hl_ffi.h they don't have to worry about the bad macro definitions like `_GUID`, but if they include hl.h directly (but without  `HL_DISABLE_LEGACY_FFI`) they still get legacy definitions as before.

We should also add `HL_DISABLE_LEGACY_FFI` to hlc templates since ffi definitions are not needed to compile hlc code and it reduces name clashes.

yuxiaomao:
Seems better indead, will do

yuxiaomao:
Hum finally I'm not sure: with your idea, when including accidentally `<hl.h>` before `<hl_ffi.h>` without `HL_DISABLE_LEGACY_FFI`, legacy definition will be present; however if the order is `<hl_ffi.h>` then `<hl.h>` it does not includes legacy definition. It can be confusing

tobil4sk:
Hm, you're right that's a good point, it's annoying when header order changes behaviour like that.

I think it would still make sense to keep the legacy definitions out of hl_ffi.h. Also, I think you shouldn't get the new definitions unless you include `hl_ffi.h`, otherwise the migration is a bit confusing.

If we reorder a few of the definitions we can make hl_ffi.h independent of hl.h. For example, I think `HL_NO_OPT` isn't really cffi related so it can just go in `hl.h`, and the `vstring` declaration as well. This way `hl_ffi.h` does not have to include `hl.h`. I think this is better as it makes the purpose of each header more concrete.

yuxiaomao:
I'm not sure how to deal with the case `HL_PRIM`/`HL_NAME` without possible redefinition problem

tobil4sk:
> I'm not sure how to deal with the case HL_PRIM/HL_NAME without possible redefinition problem

`HL_NAME` should already have `#ifndef` (or `#if !defined()`) so that the hdll can define a custom `HL_NAME`. That way it is safe for it to be in both hl.h and hl_ffi.h.

For HL_PRIM I think a similar solution using #ifndef could be used to avoid redefinition issues.

yuxiaomao:
Sorry I was not clear: I would like to avoid my use of many `ifndef HL_DISABLE_LEGACY_FFI` in `hl.h` around `DEFINE_PRIM`, and separate the `HL_PRIM` from `DEFINE_PRIM` logic. However it seems a little bit complexe due to the `!defined(HL_NAME)` defines `HL_NAME` (can't use the same condition in both hl.h and hl_ffi.h)

My current implementation, i.e. use many `ifndef HL_DISABLE_LEGACY_FFI`, does not introduce `HL_PRIM` redefinition.

yuxiaomao:
Do you have other suggestions on this?

tobil4sk:
> Do you have other suggestions on this?

I was hoping it might be possible to just repeat the logic in both `hl.h` and `hl_ffi.h`, but seems like HL_NAME makes that difficult, because HL_NAME being unset seems to be used as a way to detect if libhl is being built. Might be good to clean up the logic a bit, that might make this easier.

Ideally `hl.h` with `HL_DISABLE_LEGACY_FFI` shouldn't set `HL_PRIM`, since that is only used for ffi.

yuxiaomao:
@ncannasse please check this when you have time.

This PR add a new file `hl_ffi.h` to be used with `#define HL_DISABLE_LEGACY_FFI`, for hdll developers and avoid redefintion/undef of not prefixed defines. And to deprecated legacy ffi in the future (HL 2.0 maybe).

yuxiaomao:
I think it's a good first step to cleanup hl.h, so I'll merge it.

ncannasse:
What are we trying to solve here exactly?

yuxiaomao:
This PR solve: some macro in hl.h does not have `HL_` prefix, which sometimes collide with other macro, especially GUID (see linked issue). For compatibility the new definition is only available when `-D HL_DISABLE_LEGACY_FFI`.

By cleanup hl.h, I mean hl.h is a common header for hl, libhl, hdll, hlc. Some of the function are only meant to be used internally by hl/libhl, some other are only for libraries libhl/hdlls (not hlc).

ncannasse:
I like the idea of a single include for both HLC and embedding HL. I'm not
sure why they should be separated

tobil4sk:
`hl_ffi.h` is for defining hdll primitives, not for embedding. There isn't an overlap between hdll primitives and hlc and ideally `hlc.h` should be as minimal as possible to avoid taking up hlc symbols, which is why they are separate.

ncannasse:
I would then prefer a single hl.h with a define pre include HL_HLC set in
the .c/.h generated by HL/C that would only include what's necessary for
them.

PR #871, opened 2026-02-03
[hlc] disable warning 4756 overflow in constant arithmetic

yuxiaomao:
When assign a `Single` constant, haxe side only save it's value as `double`, and the value is cast to `float` in the VM. This produce warning C4756 overflow in constant arithmetic.

e.g

```c
r1 = 3.4028234699999998e+038;
r2 = (float)r1;
```

### DIFF
--- a/other/haxelib/templates/vs2015/__file__.vcxproj
+++ b/other/haxelib/templates/vs2015/__file__.vcxproj
@@ -101,7 +101,7 @@
       <MultiProcessorCompilation>true</MultiProcessorCompilation>
       <PrecompiledHeader>Use</PrecompiledHeader>
       <ForcedIncludeFiles>stdafx.h</ForcedIncludeFiles>
-      <DisableSpecificWarnings>4703;4100;4101;4102;4204;4221;4244;4700;4701;4702;4703;4715;4716;4723</DisableSpecificWarnings>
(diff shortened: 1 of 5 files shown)


Simn:
This looks like a code generation problem though...

yuxiaomao:
Hum, checking it again I saw that most of them is from inlined Single (probably is a haxe `Float` for the inliner) vs Single

```haxe
public static inline final SCALAR_MAX : Single = 3.40282347E+38;

var best = Math.SCALAR_MAX;
```

generates

```
OFloat r1 3.40282347E+38
OMov r2 r1
```

However I can't really remove this inline because it will break default param value and can have performance impact :<

Simn:
Looking at this:

```haxe
inline final SCALAR_MAX:Single = 3.40282347E+38;

function main() {
	var best = SCALAR_MAX;
	trace(best);
}
```

We get `[Cast:Single] [Const:Float] 3.40282347E+38` in the dump. So I guess that's the pattern genhl should detect. I just checked the other targets, gencpp turns this into `(float)(((Float)3.40282347E+38))`, which looks a bit wild but not incorrect. On JVM the constant isn't propagated for some reason, but the variable comes out as `float best = Float.MAX_VALUE` which is correct too. Funny how the decompiler detects this specific value.

The other thing that seems weird is that this example prints an empty line for me on HL. I'm not sure what I expect it to print, but probably not that.

yuxiaomao:
Indeed it seems more like a genhl's thing.
I repro the empty line print, which is weired, let me check that too.

PR #875, opened 2026-02-08
Add Opus audio format support

dtwotwo:
Opus decoding via libopusfile + libopus 1.5.2.

New primitives in fmt.c: `opus_open`, `opus_info`, `opus_read`, `opus_seek`, `opus_tell` - same pattern as OGG Vorbis.
Opus always decodes to 48kHz, supports I16 and F32.


dtwotwo:
I think it would be logical to combine Opus into one folder.
Would it be okay to provide libopus as a separate additional distribution, similar to SDL and OpenAL Soft?
Waiting feedback.

yuxiaomao:
I think it's better to publish this extension, including hashlink (hdll, haxelib) and heaps parts, in a separate repo
For your references, some works were done for sound engines: https://github.com/trethaller/hlwwise, https://github.com/yuxiaomao/hlfmod (these are native wrappers, we have some additional audio management code)

dtwotwo:
> I think it's better to publish this extension, including hashlink (hdll, haxelib) and heaps parts, in a separate repo For your references, some works were done for sound engines: https://github.com/trethaller/hlwwise, https://github.com/yuxiaomao/hlfmod (these are native wrappers, we have some additional audio management code)

Maybe yeah. A friend of mine built a library based on my pull request - [hlopus](https://github.com/gimmesoda/hlopus).
At this point, I’m unsure what to do next: should I keep maintaining this repo, or close it?

yuxiaomao:
For this Pull Request, you can close it, and put your codes in a standalone library. It can contains C and Haxe code, similar to the wwise/fmod support.
For heaps side:
- I'm not very familiar with heaps file system, but I think option exist to parse resources as FileEntry and use the bytes inside. In this case, your audio management code (decode etc) can safely stays in your own repository.
- Otherwise, you can edit your heaps fork with conditional variable `#if hlopus` to activate when your haxelib is present.
- Or, overwrite heaps code by placing a file with the same path and name in your library, and make sure `-lib hlopus` is included after? `-lib heaps`.

PR #879, opened 2026-02-16
Allow clang-cl compilation under windows

Babouchot:
This PR aims at allowing compilation of hashlink using clang-cl under windows (CLANG based compiler compatible with Microsoft's CL ABI)

### DIFF
--- a/include/libuv/src/win/atomicops-inl.h
+++ b/include/libuv/src/win/atomicops-inl.h
@@ -26,7 +26,7 @@
 
 
 /* Atomic set operation on char */
-#ifdef _MSC_VER /* MSVC */
+#if defined(_MSC_VER) && !defined(__clang__)/* MSVC */
 
 /* _InterlockedOr8 is supported by MSVC on x32 and x64. It is  slightly less */
 /* efficient than InterlockedExchange, but InterlockedExchange8 does not */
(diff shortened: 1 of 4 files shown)


Babouchot:
I added several commits to allow the remaining projects to compile with clang-cl so that the entire hashlink repo (libhl, hl and its libs) can compile with clang-cl, based on my first iteration defining HL_CLANG_CL.
Now from this version I agree that we could try to find some common rules between MSVC, MINGW and CLANG-CL but given the mixings of API and ABI that might be difficult to find a common ground, I'll try to figure something out

Apprentice-Alchemist:
Instead of a separate `HL_CLANG_CL` define I would suggest removing the `!defined(HL_LLVM)` bits and just treating clang-cl as if it was MSVC.

Apprentice-Alchemist:
Sent a PR implementing this approach, including a CI job to test it: <https://github.com/HaxeFoundation/hashlink/pull/880>

Babouchot:
Thanks for the PR, it looks like it articulates nicely with other build configs and fix some issues in this PR. 
Nonetheless, I think that this commit 40c38531ed63ac4a0adfe2036e24d1a648c4318e is still needed for project generated to C to compile under clang and GCC

yuxiaomao:
Hi, can you open a separate PR for this hlc commit?

And it would be good to add a test as in
https://github.com/HaxeFoundation/hashlink/blob/master/other/tests/header_macro_clashes/check.h

Babouchot:
Just did it in #883 😃

PR #886, opened 2026-02-22
Allow linking against system pcre2

tobil4sk:
This allows using `cmake -DWITH_SYSTEM_PCRE2=ON` or `make WITH_SYSTEM_PCRE2=1` to use the system pcre2 installation instead of the vendored version.

### DIFF
--- a/src/std/regexp.c
+++ b/src/std/regexp.c
@@ -28,9 +28,9 @@ typedef struct _ereg ereg;
 struct _ereg {
 	void (*finalize)( ereg * );
 	/* The compiled regex code */
-	pcre2_code *regex;
+	pcre2_code_16 *regex;
 	/* Pointer to the allocated memory for match data */
-	pcre2_match_data *match_data;
+	pcre2_match_data_16 *match_data;
 	/* Number of capture groups */
 	int n_groups;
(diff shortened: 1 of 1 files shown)


yuxiaomao:
I'm not sure why there is the change from `pcre2_code` to `pcre2_code_16` ? As `PCRE2_CODE_UNIT_WIDTH=16` is always defined.

tobil4sk:
> I'm not sure why there is the change from pcre2_code to pcre2_code_16 ? As PCRE2_CODE_UNIT_WIDTH=16 is always defined.

It's not needed, but I get errors in my IDE sometimes because it doesn't know that `PCRE2_CODE_UNIT_WIDTH=16` is set in the build system (outside of the source file). I also thought it might be better to be explicit, since it wouldn't work if compiled with anything other than 16 bit. However, I can revert it if you prefer.

yuxiaomao:
Yes IDE often requires additional setups.
I didn't remark that it only works in 16 bit mode, the code in regexp.c seems quite generic - but I now see `vbyte* str` is casted to `PCRE2_SPTR` so it probably won't work if that isn't 16.

PR #888, opened 2026-02-23
[ci] Enable some disabled build configurations

tobil4sk:
- Mac x86 make - this was disabled in #821 because the intel runner was removed, but there is now `macos-15-intel`
- Windows mingw 32 bit - left out in #831
- Linux arm64 - there is now an arm64 linux runner so we can use it
- Linux x86 32 bit - this works fine with sdl.hdll disabled (32 bit graphics packages are not available anymore)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -257,7 +257,7 @@ if(BUILD_TESTING)
     )
 
     if(CMAKE_SIZEOF_VOID_P EQUAL 4)
-        SET(HAXE_FLAGS -D hl-legacy32)
+        SET(HAXE_HL_FLAGS -D hl-legacy32)
     endif()
 
     if(MINGW)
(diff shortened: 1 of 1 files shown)


yuxiaomao:
I'm happy to see Mac x86 and Linux arm64 in CI, but I'm not sure if it worth to enable much 32-bit build (not enough tests and is known broken for more complexe usage).

tobil4sk:
I think it makes sense to have mingw32 since windows still has 32 bit builds for cmake and VS.

For linux 32 bit, it still passes all the haxe unit tests when building with hlc (even without -D hl_legacy32). Part of the reason I enabled it is to figure out what the state is.

yuxiaomao:
Ok, I was thinking about hl jit broken (it's probably not that far from working) instead of hl c.
I'm good with this PR. It'll probably have some minor conflicts with the makefile PR, I perfer waiting for that PR first because it seems easier in this order.

yuxiaomao:
Maybe it can be useful for other PRs (I saw some of the changes in AArch64 PR), so let me merge this now

PR #890, opened 2026-02-23
[make] Clean up variables and rules

tobil4sk:
Uses more standard variable naming (CFLAGS, CPPFLAGS, LDFLAGS, see: https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html), which allows using builtin make rules for compiling source files.

Instead of defining a compilation rule for each hdll, there is now a generic pattern for hdll files and each specific hdll just has to define its dependency objects and any additional flags it needs (using: https://www.gnu.org/software/make/manual/html_node/Target_002dspecific.html).

I've also made most of the rules into non-PHONY rules, which means make can correctly detect when a file needs to be rebuilt instead of rebuilding it every time.

Closes #571.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,6 @@ ARCH ?= $(shell uname -m)
 
 CFLAGS = -Wall -O3 -std=c11 -fvisibility=hidden
 CPPFLAGS = -I src
-LDFLAGS = -fvisibility=hidden
 LIBHL_LDFLAGS =
 LIBHL_LDLIBS = -lm -lpthread
 USE_LIBHL_LDFLAGS =
@@ -151,7 +150,6 @@ endif
 
 ifneq (, $(findstring MINGW64, $(UNAME)))
 CFLAGS += -municode
-LDFLAGS += -municode
 LIBHL_LDLIBS += -lws2_32 -lwsock32
 hl_LDLIBS = -lm
 hlc_LDLIBS = -ldbghelp
@@ -170,7 +168,6 @@ BREW_OPENAL_PREFIX := $(shell brew --prefix openal-soft)
 BREW_SDL_PREFIX := $(shell brew --prefix sdl2)
 
 CFLAGS += -arch $(ARCH)
-LDFLAGS += -arch $(ARCH)
 CPPFLAGS += -I include -I $(BREW_PREFIX)/include
 
 SDL_CPPFLAGS = -I $(BREW_SDL_PREFIX)/include/SDL2 -DGL_SILENCE_DEPRECATION
@@ -179,7 +176,6 @@ OPENAL_CPPFLAGS = -I $(BREW_OPENAL_PREFIX)/include -Dopenal_soft
 ifdef OSX_SDK
 ISYSROOT = $(shell xcrun --sdk macosx$(OSX_SDK) --show-sdk-path)
 CFLAGS += -isysroot $(ISYSROOT)
-LDFLAGS += -isysroot $(ISYSROOT)
 endif
 
 LDFLAGS += -L$(BREW_PREFIX)/lib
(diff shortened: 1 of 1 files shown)


yuxiaomao:
I didn't fully understand everything but seems good (and I learned things like `$(LINK.c)` and suffixes rules).

My only question is for `.SUFFIXES` and the removal of `.c.o` rule, in the documentation
https://www.gnu.org/software/make/manual/html_node/Suffix-Rules.html

> Suffix rules are obsolete because pattern rules are more general and clearer. They are supported in GNU make for compatibility with old makefiles.

Will it be better to keep .c.o rule (and add .cpp.o)?

I also feel it'll be easier to understand if `$(LINK.c)` / `$(COMPILE.c)` are replaced with theirs actual definition (it was not very clear when searching on web, then I learned that `make -p` can print the definition), but that's probably just personal.

tobil4sk:
> My only question is for .SUFFIXES and the removal of .c.o rule, in the documentation

I didn't notice these are related to be honest, but yes I guess `.SUFFIXES` should be removed. For `.c.o`, that is the old suffix rule it is talking about, so it shouldn't be added back. We should rely on pattern rules instead, and Make has default rules for c and c++ files which work as long as we use the standard variables like CFLAGS, CPPFLAGS, etc.

> I also feel it'll be easier to understand if $(LINK.c) / $(COMPILE.c) are replaced with theirs actual definition (it was not very clear when searching on web, then I learned that make -p can print the definition), but that's probably just personal.

That's a good point. I was already a bit unsure about using those because I couldn't tell if they were internal or not, but if searching on the web doesn't explain it then it is probably a good sign to not use it.

tobil4sk:
Seems like `COMPILE.c` and `LINK.c` are actually documented:

https://www.gnu.org/software/make/manual/html_node/Catalogue-of-Rules.html

> make follows the convention that the rule to compile a .x source file uses the variable COMPILE.x. Similarly, the rule to produce an executable from a .x file uses LINK.x; and the rule to preprocess a .x file uses PREPROCESS.x. 

I wonder if it would be better to follow the convention in that case and use them. There are quite a few variables that make up the `COMPILE.c` and `LINK.c` definition, and I think using them might be more portable for systems with slight differences from what we might hardcode. It means we have to define/maintain less ourselves and can rely on make to deal with the low level details of how to compile files on the given platform. What do you think?

> > My only question is for .SUFFIXES and the removal of .c.o rule, in the documentation

>I didn't notice these are related to be honest, but yes I guess .SUFFIXES should be removed.

It looks like there is still a use, to restrict predefined rules added by make: https://www.gnu.org/software/make/manual/html_node/Catalogue-of-Rules.html
> If you modify the suffix list, the only predefined suffix rules in effect will be those named by one or two of the suffixes that are on the list you specify; rules whose suffixes fail to be on the list are disabled.

I have changed it to also clear `.SUFFIXES` and added a comment to explain. I think it is good because it makes it explicit what predefined behaviour we rely on.

yuxiaomao:
> Seems like COMPILE.c and LINK.c are actually documented:

Yes, they have documentation, and I agree with you that using them "might be more portable for systems with slight differences". But searching google naivly with COMPILE.c or LINK.c with some basic keywords does not give the related documentation, which is why I'm confused at the first place. That why I feel it'll be easier to understand if we could use their definition and all are explicit inside the Makefile.

> I have changed it to also clear .SUFFIXES and added a comment to explain. I think it is good because it makes it explicit what predefined behaviour we rely on.

I feel similar for .c.o and .cpp.o rules hidden inside .SUFFIX, but finding the doc was much easier so I'm fine with the current state.

tobil4sk:
>  But searching google naivly with COMPILE.c or LINK.c with some basic keywords does not give the related documentation, which is why I'm confused at the first place. That why I feel it'll be easier to understand if we could use their definition and all are explicit inside the Makefile.

I have replaced them with the definitions. It also encourages proper use of `LDFLAGS` so I think it's a bit cleaner now. (I also noticed that the previous version still used `-o` directly anyway).

PR #902, opened 2026-03-05
Use native Windows unwinding.

### LINKED ISSUE #892: Exception stack misses actual throw position
```haxe
import haxe.Exception;

function f1() {
	throw new Exception("f1");
}

function f2() {
	f1();
}

function main() {
	try {
		f2();
	} catch(e:Exception) {
		for (item in e.stack.asArray()) {
			trace(item);
		}
	}
}
```

HL:

```
source/Main.hx:16: FilePos(Method(_Main.$Main_Fields_,f2),source/Main.hx,8,null)
source/Main.hx:16: FilePos(Method(_Main.$Main_Fields_,main),source/Main.hx,13,null)
source/Main.hx:16: FilePos(Method(_Main.$Main_Fields_,main),source/Main.hx,12,null)
```

All other targets (that I've tried) have the original `FilePos(Method(_Main.Main_Fields_,f1),source/Main.hx,4,null)` as part of the stack.

I'm also just now seeing the line 12 entry, which seems to leak some internal exception handling mechanism because that's on the `try`.

Apprentice-Alchemist:
- generate Windows unwind info
- remove custom longjump hack
- implement native unwinder to improve backtrace reliability

Closes https://github.com/HaxeFoundation/hashlink/issues/892

### DIFF
--- a/src/hlmodule.h
+++ b/src/hlmodule.h
@@ -20,6 +20,7 @@
  * DEALINGS IN THE SOFTWARE.
  */
 #include <hl.h>
+#include <hlsystem.h>
 #include "opcodes.h"
 
 typedef struct {
@@ -115,6 +116,10 @@ typedef struct {
 	int *functions_indexes;
 } hl_code_hash;
 
+#if defined(HL_64) && defined(HL_WIN)
+#define WIN64_UNWIND_TABLES
+#endif
+
 typedef struct {
 	hl_code *code;
 	int codesize;
@@ -128,6 +133,9 @@ typedef struct {
 	hl_debug_infos *jit_debug;
 	jit_ctx *jit_ctx;
 	hl_module_context ctx;
+#ifdef WIN64_UNWIND_TABLES
+	PRUNTIME_FUNCTION unwind_table;
+#endif
 } hl_module;
 
 hl_code *hl_code_read( const unsigned char *data, int size, char **error_msg );
--- a/src/jit.c
+++ b/src/jit.c
@@ -266,10 +266,6 @@ static const int RCPU_SCRATCH_REGS[] = { Eax, Ecx, Edx };
 
 #define BREAK()		B(0xCC)
 
-#if defined(HL_64) && defined(HL_VCC)
-#	define JIT_CUSTOM_LONGJUMP
-#endif
-
 static preg _unused = { RUNUSED, 0, 0, NULL };
 static preg *UNUSED = &_unused;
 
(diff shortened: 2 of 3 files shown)


Simn:
```
C:\git\hxcoro\callstack-tests>haxe build-hl.hxml
PASS asyncscope
PASS awaitcontinuation
PASS awaittask
PASS cancellation
PASS catchrethrow
PASS directthrow
PASS foobarbaz
PASS implicitlazy
PASS lazytask
PASS nestedplainthrow
PASS scopetask
PASS siblingcancellation
PASS supervisortask
PASS timeout
PASS toprecursion

Results: 15 passed, 0 failed
```

Humanity wins!

yuxiaomao:
Something strange happens with this PR:

```haxe
function main() {
	var str = Std.string(1.185);
	Sys.println(str);
}
```

Expected `1.185`, got `2.46228694691432e+77`

yuxiaomao:
Haxe unit test isn't very happy too, run locally and it gives 368 failures

```
utest/ui/text/PrintReport.hx:52:
assertations: 11628
successes: 11258
errors: 2
failures: 368
warnings: 0
execution time: 0
```

Simn:
Works for me now! The hxcoro tests pass too. They were actually passing before the update as well but printed `execution time: NaN` at the end, which is why I didn't notice the problem.

Simn:
(Side note: this means that the CI here doesn't run the Haxe unit tests, which it should like hxcpp does.)

yuxiaomao:
I can't fully understand it, but old comment already said it's better to use RtlAddFunctionTable, and unit tests and Simon are happy, so seems good to me.

> // Another more reliable way of handling this would be to use RtlAddFunctionTable
> // but this would require complex creation of unwind info

Simn:
@Apprentice-Alchemist Could you briefly explain why the first version had this unintended side-effect? From what I've seen it affected floating point values which seems quite surprising.

Apprentice-Alchemist:
> @Apprentice-Alchemist Could you briefly explain why the first version had this unintended side-effect? From what I've seen it affected floating point values which seems quite surprising.

The JIT stores floating point values at the start of the code buffer and expect them to start at offset zero.

The first version was storing the unwind data at offset zero, so the wrong floating point values where being loaded.

yuxiaomao:
On Xbox Series, `RtlAddFunctionTable` and `RtlDeleteFunctionTable` do not exist (only `RtlUnwind`).

Is that still a good choice to restore old long_jump for `HL_XBS` platform?
Also I'm not sure if I should keep 1 argument if using old long jump, for the `On Win64 setjmp actually takes two arguments` part.

yuxiaomao:
Note: native longjmp cause crash on very recent CPU (Intel CET), custom longjmp for throw is restored for Windows in https://github.com/HaxeFoundation/hashlink/commit/781960a5daca32ad6d5cea87b255fe8b5872551e

yuxiaomao:
On `hl.UI.Sentinel` timeout, `module_capture_stack` has access violation (wrong simulated stack ?). Before this commit, the repro below calls the callback and access violation later in the debugger (needs to change ui_win and remove skip debugger condition) - but continue to "work" if no debugger, as if we choose "continue". I'm still trying to understand what's going on ^^'

```
Exception thrown at 0x00007FF81E5DEE30 (ntdll.dll) in hl.exe: 0xC0000005: Access violation reading location 0xFFFFFFFFFFFFFFFF.

Access violation
Called from haxe.$Exception.__constructor__ (haxe/Exception.hx line 46)
Called from haxe.$ValueException.__constructor__ (haxe/ValueException.hx line 24)
Called from haxe.$Exception.thrown (haxe/Exception.hx line 33)
```

repro with:
```haxe
class Main extends hxd.App {
	override function update(dt:Float) {
		if( hxd.Key.isDown(hxd.Key.SPACE) )
			stuck();
	}

	function stuck() {
		trace("In stuck");
		while( true ) {};
	}

	static function main() {
		new Main();
	}
}
```

PR #927, opened 2026-04-27
profile.c: implement SP/PC registers for ARM64 in get_thread_stackptr

HomuHomu833:
I encountered this issue while compiling in Lime, so I would like to submit a PR to fix this issue.

### DIFF
--- a/src/profile.c
+++ b/src/profile.c
@@ -138,7 +138,10 @@ static void *get_thread_stackptr( thread_handle *t, void **eip ) {
 	CONTEXT c;
 	c.ContextFlags = CONTEXT_CONTROL;
 	if( !GetThreadContext(t->h,&c) ) return NULL;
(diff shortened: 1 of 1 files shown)


tobil4sk:
I think `profile.c` is part of the `hl` binary, which is currently not possible to compile on arm64 anyway (it would require #895).

HomuHomu833:
indeed but I can compile hl for aarch64 and even use Lime to build FNF. I think Lime only suffers from this

image
image

feel free to close the PR if needed I guess then.

tobil4sk:
Lime should be modified to omit the `hl` binary if targetting arm64 and to only support hlc for arm64. Once proper arm64 support is added to hashlink jit then this issue will be solved.

HomuHomu833:
will wait https://github.com/HaxeFoundation/hashlink/pull/895

PR #929, opened 2026-05-01
Fix crash when bogus string input is provided to hl_date_from_string

sebbernery:
Fix crash when bogus string input is provided to hl_date_from_string or format digest function.

Added a new function in hl.h, hl_to_utf8_len, same as hl_to_utf8 but with a len parameter Updated hl_utf16_to_utf8 to replace '!= end' by '< end' because c can be incremented two times in the loop and it may be an issue with invalid input.

Here is a reproductible example : 

[segfault_fixed_by_utf8_len.txt](https://github.com/user-attachments/files/27279516/segfault_fixed_by_utf8_len.txt)

```haxe
class Main {
    static public function main() {
        final input = Sys.stdin().readLine();
        try {
            var date = Date.fromString(input);
            trace(date);
        } catch (e:haxe.Exception) {
        }
    }
}
```

```
$ haxe -main Main -hl out.hl
$ hl out.hl < segfault_fixed_by_utf8_len.txt
SIGNAL 11[Segmentation fault]
$Date.fromString(/usr/share/haxe/std/hl/_std/Date.hx:157)
$ParseDate.main(ParseDate.hx:5)
.init(?:1)
[1]    2231895 segmentation fault
```

Here is the reproductible example for Shadigest (with the same input):
```
import haxe.crypto.Sha1;

class Main {
    static function main() {
        final input = Sys.stdin().readLine();
        try {
            trace(Sha1.encode(input));
        } catch (e:haxe.Exception) {
        }
    }
}
```

```
$ haxe -main Main -hl out.hl
$ hl out.hl < segfault_fixed_by_utf8_len.txt
SIGNAL 11[Segmentation fault]
haxe.crypto.$Sha1.encode(/usr/share/haxe/std/hl/_std/haxe/crypto/Sha1.hx:28)
$Main.main(Main.hx:8)
.init(?:1)
[1]    2671945 segmentation fault
```
I found this issue with AFL++ by trying fuzzing on some Haxe std function. Don't hesitate to ask if you need more details (I kept some notes during the debugging process, but it's in french). 

I didn't used any LLM in the entire process from writting Haxe script or fixes or this PR.
Have a nice day !

### DIFF
--- a/libs/fmt/fmt.c
+++ b/libs/fmt/fmt.c
@@ -750,7 +750,7 @@ static void md5_finish( md5_context *ctx, uint8 digest[16] ) {
 
 HL_PRIM void HL_NAME(digest)( vbyte *out, vbyte *in, int length, int format ) {
 	if( format & 256 ) {
-		in = (vbyte*)hl_to_utf8((uchar*)in);
+		in = (vbyte*)hl_to_utf8_len((uchar*)in, length);
 		length = (int)strlen((char*)in);
 	}
 	hl_blocking(true);
--- a/src/hl.h
+++ b/src/hl.h
@@ -650,6 +650,7 @@ HL_API vbyte *hl_copy_bytes( const vbyte *byte, int size );
 HL_API int hl_utf8_length( const vbyte *s, int pos );
 HL_API int hl_from_utf8( uchar *out, int outLen, const char *str );
 HL_API char *hl_to_utf8( const uchar *bytes );
+HL_API char *hl_to_utf8_len( const uchar *bytes, int len );
 HL_API uchar *hl_to_utf16( const char *str );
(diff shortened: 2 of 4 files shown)


sebbernery:
I just noted that my fix is incorrect as it can raise a segfault when there was no problem before with the date parsing, I think it comes from this : https://github.com/HaxeFoundation/haxe/blob/development/std/hl/_std/Date.hx#L157 . The length of the string given is multiplied by two, so I think it's not coherent with the real size of the buffer. I don't get why. 

I switched the PR in draft mode.

yuxiaomao:
Note also: on Windows, the try-catch catched the access violation, and it won't access violation if I trace input before `Date.fromString`

PR #932, opened 2026-05-11
AArch64 backend for the IR-based JIT

bmdhacks:
Adds an AArch64 codegen backend that plugs into the new IR pipeline (jit.c -> jit_regs.c -> jit_aarch64.c), mirroring the role of jit_x86_64.c. Targets Linux and macOS ARM64 (AAPCS64; X15/X16/X17 reserved as backend temporaries, 16-byte internal push stride, native overflow arguments at the platform ABI).

Backend surface:
  hl_jit_init_regs       AAPCS64 register classes (X0..X14 scratch with X5
                         remapped to logical slot 32, X19..X28 persist;
                         V0..V7 and V16..V28 scratch, V8..V15 persist,
                         V29..V31 reserved)
  hl_codegen_init        module preamble: null-access stubs, c2hl/hl2c
                         trampolines
  hl_codegen_function    IR einstr stream -> AArch64 machine code
  hl_codegen_flush_consts ADRP+LDR patching for the per-module pool
                         (distinguishes LDR Xt/Dt/St for correct imm12
                         scaling)
  hl_codegen_final       BL relocations, jump-table absolute fixups

Encoding layer (jit_aarch64_emit.c/h) supplies the instruction encoders shared with the trampolines; floating conditional moves use FCSEL and parallel-move cycles use XCHG/CXCHG.

regs_config gains a stack_arg_size field so the IR's stack-argument accounting agrees with the backend's 16-byte-per-push convention required by AAPCS64 stack alignment.

module.c: on AArch64, walk the X29 frame-pointer chain in module_capture_stack instead of the heuristic stack scanner. The scanner produces false-positive frames from callee-saved STP X19,X20 spills that look like (stack_addr, code_addr) pairs. PUSH_ADDR emits output-relative ADRP+ADD fixups so null-access frames resolve to the correct source line.

Build: CMake selects src/jit_aarch64.c when the target processor is aarch64/arm64; the Makefile gains an ARCH=arm64 branch. CI runs the ARM64 JIT tests.

macOS arm64 code memory uses MAP_JIT and pthread_jit_write_protect_np.

Validated against unit.hl with all tests passing.

### DIFF
--- a/src/jit.c
+++ b/src/jit.c
@@ -332,6 +332,7 @@ void *hl_jit_code( jit_ctx *ctx, hl_module *m, int *codesize, hl_debug_infos **d
 	ctx->final_code = code;
 	hl_emit_final(ctx);
 	hl_codegen_final(ctx);
+	hl_flush_executable_memory(code, size);
 	arg_reg_count = ctx->cfg.regs.nargs;
 	arg_fp_count = ctx->cfg.floats.nargs;
 	call_jit_c2hl = ctx->final_code + ctx->code_funs.c2hl;
--- a/src/jit.c
+++ b/src/jit.c
@@ -357,6 +357,7 @@ void *hl_jit_code( jit_ctx *ctx, hl_module *m, int *codesize, hl_debug_infos **d
 	ctx->final_code = code;
 	hl_emit_final(ctx);
 	hl_codegen_final(ctx);
+	hl_flush_executable_memory(code, size);
 	arg_reg_count = ctx->cfg.regs.nargs;
 	arg_fp_count = ctx->cfg.floats.nargs;
 	call_jit_c2hl = ctx->final_code + ctx->code_funs.c2hl;
--- a/src/jit.h
+++ b/src/jit.h
@@ -190,6 +190,18 @@ typedef struct {
 	ereg *arg;
 } reg_config;
 
+// HL→native stack-arg layout. HL→HL calls always use the HL convention
+// (cfg.stack_arg_size stride, PUSH-based). HL→native calls follow the
+// platform C ABI, which on AArch64 differs from HL's convention:
+//   - Linux/Windows AArch64 (AAPCS64): each stack arg occupies an 8-byte slot.
+//   - Apple ARM64: stack args are packed at their natural size with
+//     natural alignment.
+typedef enum {
+	NATIVE_STACK_LAYOUT_HL = 0,        // default: same as HL stride (x86_64, etc.)
+	NATIVE_STACK_LAYOUT_AAPCS64,       // 8-byte slots (Linux/Win AArch64 native)
+	NATIVE_STACK_LAYOUT_APPLE_ARM64,   // natural-size packed (Apple ARM64 native)
+} native_stack_layout_kind;
+
 typedef struct {
 	reg_config regs;
 	reg_config floats;
@@ -201,6 +213,8 @@ typedef struct {
 	// move SP by 16 bytes to keep SP 16-byte aligned (any [SP, ...] access
 	// with a misaligned SP traps under EL0).
 	int stack_arg_size;
+	// Layout for HL→native (CALL_PTR) stack args. See native_stack_layout_kind.
+	int native_stack_layout;
 	int debug_prefix_size;
 	ereg req_bit_shifts;
 	ereg req_div_a;
--- a/src/jit.h
+++ b/src/jit.h
@@ -196,6 +196,11 @@ typedef struct {
 	ereg stack_reg;
 	ereg stack_pos;
 	int stack_align;
+	// Minimum bytes consumed by each stack argument. Defaults to HL_WSIZE
+	// when 0. Backends like AArch64 set this to 16 because each PUSH must
+	// move SP by 16 bytes to keep SP 16-byte aligned (any [SP, ...] access
+	// with a misaligned SP traps under EL0).
+	int stack_arg_size;
 	int debug_prefix_size;
 	ereg req_bit_shifts;
 	ereg req_div_a;
--- a/src/jit.h
+++ b/src/jit.h
@@ -197,12 +197,31 @@ typedef struct {
 	ereg *arg;
 } reg_config;
 
+// HL->native stack-arg layout. HL->HL calls always use the HL convention
+// (cfg.stack_arg_size stride, PUSH-based). HL->native calls follow the
+// platform C ABI, which on AArch64 differs from HL's convention:
+//   - Linux/Windows AArch64 (AAPCS64): each stack arg occupies an 8-byte slot.
+//   - Apple ARM64: stack args are packed at their natural size with
+//     natural alignment.
+typedef enum {
+	NATIVE_STACK_LAYOUT_HL = 0,        // default: same as HL stride (x86_64, etc.)
+	NATIVE_STACK_LAYOUT_AAPCS64,       // 8-byte slots (Linux/Win AArch64 native)
+	NATIVE_STACK_LAYOUT_APPLE_ARM64,   // natural-size packed (Apple ARM64 native)
+} native_stack_layout_kind;
+
 typedef struct {
 	reg_config regs;
 	reg_config floats;
 	ereg stack_reg;
 	ereg stack_pos;
 	int stack_align;
+	// Minimum bytes consumed by each stack argument. Defaults to HL_WSIZE
+	// when 0. Backends like AArch64 set this to 16 because each PUSH must
+	// move SP by 16 bytes to keep SP 16-byte aligned (any [SP, ...] access
+	// with a misaligned SP traps under EL0).
+	int stack_arg_size;
+	// Layout for HL->native (CALL_PTR) stack args. See native_stack_layout_kind.
+	int native_stack_layout;
 	int debug_prefix_size;
 	ereg req_bit_shifts;
 	ereg req_div_a;
--- a/src/jit_regs.c
+++ b/src/jit_regs.c
@@ -102,6 +102,41 @@ static int get_stack_size( regs_ctx *ctx, emit_mode m ) {
 	return size;
 }
 
+// Compute the native-ABI stack-arg layout for an HL→native call.
+// Fills offsets[k] with the byte offset (from the post-allocation SP) where
+// arg k lives on the stack, or -1 if arg k is passed in a register.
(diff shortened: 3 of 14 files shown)


src/jit.c
```diff
@@ -325,12 +334,17 @@ void *hl_jit_code( jit_ctx *ctx, hl_module *m, int *codesize, hl_debug_infos **d
...
 	memcpy(code,ctx->output,size);
 	*codesize = size;
 	*debug = m->jit_debug;
 	ctx->final_code = code;
 	hl_emit_final(ctx);
 	hl_codegen_final(ctx);
+#if defined(__GNUC__) || defined(__clang__)
+	__builtin___clear_cache((char*)code, (char*)code + size);
+#endif
+	HL_JIT_WRITE_PROTECT(true);
```
ncannasse: I think this should be done with a call to a new `hl_flush_executable_memory(code,size);` so we don't need these HL_WRITE_PROTECT and clear cache here.

src/jit_regs.c
```diff
@@ -512,11 +554,16 @@ static void flush_movs( regs_ctx *ctx, bool cond ) {
...
+			// and is unaffected by the XCHG.)
 			for(int k=0;k<size;k+=3) {
-				if( int_arr_get(movs,k) == to )
-					movs.values[k] = from;
-				else if( int_arr_get(movs,k) == from )
-					movs.values[k] = to;
+				if( int_arr_get(movs,k+1) == to )
+					movs.values[k+1] = from;
+				else if( int_arr_get(movs,k+1) == from )
+					movs.values[k+1] = to;
```
ncannasse: fixed in IR branch

src/jit.c
bmdhacks: I added hl_flush_executable_memory(void *code, int size)

For the apple hardware-enforced write-xor-execute policy I mark the region as writable on alloc and then executable when we flush. Now jit.c has no platform ifdefs anymore.

bmdhacks:
rebased to latest hl2_ir code

tritao:
@bmdhacks Can you rebase this again?

bmdhacks:
> @bmdhacks Can you rebase this again?

Working on this now!

bmdhacks:
Updated to latest.  All unit tests pass and it runs Dead Cells on asahi linux on my M2 mac.
Screenshot From 2026-09-11 21-38-01

bmdhacks:
BTW, @ncannasse I avoided making any changes to the IR since I wanted this to be as drop-in as possible to make it easyish for you to accept (I know it's still a big giant wall of code).  But there are some more optimizations I could do like using NEON for some ops if you'd be up for a later PR with some IR additions/alterations.

Also thanks for your time and consideration and all the work.  Your contributions to gaming are amazing.

src/jit.h
```diff
@@ -197,12 +197,31 @@ typedef struct {
...
 	reg_config regs;
 	reg_config floats;
 	ereg stack_reg;
 	ereg stack_pos;
 	int stack_align;
+	// Minimum bytes consumed by each stack argument. Defaults to HL_WSIZE
+	// when 0. Backends like AArch64 set this to 16 because each PUSH must
+	// move SP by 16 bytes to keep SP 16-byte aligned (any [SP, ...] access
+	// with a misaligned SP traps under EL0).
+	int stack_arg_size;
```
ncannasse, line 222: min_stack_args_size ?

src/jit.h
```diff
@@ -197,12 +197,31 @@ typedef struct {
...
+// (cfg.stack_arg_size stride, PUSH-based). HL->native calls follow the
+// platform C ABI, which on AArch64 differs from HL's convention:
+//   - Linux/Windows AArch64 (AAPCS64): each stack arg occupies an 8-byte slot.
+//   - Apple ARM64: stack args are packed at their natural size with
+//     natural alignment.
+typedef enum {
+	NATIVE_STACK_LAYOUT_HL = 0,        // default: same as HL stride (x86_64, etc.)
+	NATIVE_STACK_LAYOUT_AAPCS64,       // 8-byte slots (Linux/Win AArch64 native)
+	NATIVE_STACK_LAYOUT_APPLE_ARM64,   // natural-size packed (Apple ARM64 native)
+} native_stack_layout_kind;
```
ncannasse, line 210: maybe the constants names should be more about the algorith than the platform (for example ALIGN_64 for AAPPCS64) and shorter like LAYOUT_ALIGN_64 , for "HL" that would be LAYOUT_ALIGN_SIZE , I'm not sure how's apple is different than intel here.
If it comes down to just minimum alignment then it should just be a int for stack_align_bytes

src/jit.h
```diff
@@ -197,12 +197,31 @@ typedef struct {
...
 	ereg stack_reg;
 	ereg stack_pos;
 	int stack_align;
+	// Minimum bytes consumed by each stack argument. Defaults to HL_WSIZE
+	// when 0. Backends like AArch64 set this to 16 because each PUSH must
+	// move SP by 16 bytes to keep SP 16-byte aligned (any [SP, ...] access
+	// with a misaligned SP traps under EL0).
+	int stack_arg_size;
+	// Layout for HL->native (CALL_PTR) stack args. See native_stack_layout_kind.
+	int native_stack_layout;
```
ncannasse, line 224: int -> native_stack_layout_kind ?

src/jit_regs.c
```diff
@@ -102,12 +102,49 @@ static ereg get_call_reg( regs_ctx *ctx, call_regs regs, emit_mode m ) {
...
 	if( size < HL_WSIZE ) size = HL_WSIZE;
+	int min = ctx->jit->cfg.stack_arg_size;
+	if( min && size < min ) size = min;
 	return size;
 }
 
+// Compute the native-ABI stack-arg layout for an HL->native call.
+// Fills offsets[k] with the byte offset (from the post-allocation SP) where
+// arg k lives on the stack, or -1 if arg k is passed in a register.
+// Returns the total stack frame size in bytes (rounded up to 16-byte align).
```
ncannasse, line 116: please tell ai to only keep the comments when they are useful :)

src/jit_regs.c
```diff
@@ -102,12 +102,49 @@ static ereg get_call_reg( regs_ctx *ctx, call_regs regs, emit_mode m ) {
...
+			offsets[k] = offs;
+			offs += size;
+		} else {
+			// AAPCS64 (Linux/Win) and other AArch64 ABIs: 8-byte slots.
+			offsets[k] = offs;
+			offs += HL_WSIZE;
+		}
+	}
+	// Pad total frame to 16-byte alignment (AAPCS64 SP alignment requirement).
+	if( offs & 15 ) offs = (offs + 15) & ~15;
```
ncannasse, line 144: it is ARCH64 specific ?

src/jit_regs.c
```diff
@@ -483,8 +520,13 @@ static void regs_assign_regs( regs_ctx *ctx ) {
 			}
 		}
 		if( IS_NULL(r) || IS_WINCALL64 ) {
-			// use existing stack storage
-			v->stack_pos = (args_count++ + 2) * HL_WSIZE;
+			// use existing stack storage. Stride must match the caller's
+			// get_stack_size() so HL->HL stack args land where the callee
+			// reads them. AArch64 uses 16-byte stride (cfg.stack_arg_size)
+			// because each PUSH moves SP by 16 to keep SP 16-byte aligned;
+			// x86_64 (and Win64 shadow stack) uses HL_WSIZE.
```
ncannasse, line 527: we don't want any comment about specific architecture here, the whole point is to have everything arch specific in cfg

src/jit_regs.c
```diff
@@ -599,7 +641,8 @@ static void regs_assign_regs( regs_ctx *ctx ) {
 	}
 	// assign stack regs
 	int nvalues = jit->value_count + jit->phi_count;
-	int persists_size = (ctx->persists_uses[0] + ctx->persists_uses[1]) * 8;
+	int push_unit = jit->cfg.stack_arg_size ? jit->cfg.stack_arg_size : HL_WSIZE;
```
ncannasse, line 644: if we always test it for 0 then better set it to HL_WSIZE in x86_64 and not do any test here

src/jit_regs.c
```diff
@@ -740,20 +784,46 @@ static void regs_emit_instrs( regs_ctx *ctx ) {
...
 			int stack_args = 0;
 			int stack_bits = 0;
+			// HL->native calls (CALL_PTR) on backends with a non-default
+			// native_stack_layout (currently only AArch64) follow the
+			// platform C ABI rather than HL's internal convention. Compute
+			// the per-arg offsets up front; we emit one STACK_OFFS to allocate
+			// the frame and individual STORE instructions instead of the
+			// PUSH-based path used for HL->HL.
+			bool native_layout = (e.op == CALL_PTR && jit->cfg.native_stack_layout != NATIVE_STACK_LAYOUT_HL);
+			int native_offsets[256];
```
ncannasse, line 794: we already have MAX_CALL_ARGS

src/module.c
```diff
@@ -252,6 +252,72 @@ static int module_capture_stack( void **stack, int size ) {
...
+				stack[count] = addr;
+			}
+			count++;
+		}
+		prev_helper = (kind == 1);
+		if( next_fp == NULL || next_fp <= (void *)fp || next_fp >= stack_top )
+			break;
+		fp = (void **)next_fp;
+	}
+	return count;
```
ncannasse, line 320: I think this could be better written with a  hl_module *hl_resolve_module( void * addr ) then calling that instead of doing loops.

src/jit.h
bmdhacks: Ok, I made it the int argument: `min_native_stack_args_size`:
8=linux/windows aarch64
1=apple (apple packs the stack with actual size of the arg)
0=x86 where native calls use the push path

src/jit_regs.c
bmdhacks: sorry bout that

src/jit_regs.c
bmdhacks: normal HL calls with more than MAX_ARGS could use this path too so I just make it go through the arg list twice to calculate their offsets and stack size

bmdhacks:
I addressed your comments, thanks for the review.  The last commit is the most important where on arm HL calls now pass stack arguments using the native ABI layout, as they do on x86.
I had to make callback_c2hl and callback_hl2c a bit more flexible to stack layout.

PR #936, opened 2026-05-15
Fix macOS Homebrew package install in CI

dtwotwo:
Fixes macOS CI failures caused by Homebrew upgrading installed dependents while installing required packages.

Sets HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=1 for macOS brew installs so unrelated runner packages like python@3.13 are not upgraded during the build.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -167,7 +167,7 @@ jobs:
           darwin)
             if [ "${{matrix.build_system}}" != "cmake" ]; then
               brew update
-              brew bundle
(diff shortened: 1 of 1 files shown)


kLabz:
Also note that there are `brew install` commands run from `tests/runci/targets/{Lua,Php,Python}.hx` too

Edit: eh, didn't notice that was hashlink repo.. xD

dtwotwo:
> Also note that there are `brew install` commands run from `tests/runci/targets/{Lua,Php,Python}.hx` too

Okay, I added the env var at the haxe-test-suite job level so brew installs launched by RunCi inherit it too.

PR #952, opened 2026-06-28
libs/sdl: build on Android / GL ES (glColorMaski stub + SDL_SetMainReady)

Pign:
Two small fixes to build the SDL/GL bindings on Android arm64 (GL ES):

- `gl.c`: stub `glColorMaski` to `ES_NOT_SUPPORTED` — GL ES has no per-MRT colour mask, so the call can't be expressed there. Keeps the desktop path untouched.
- `sdl.c`: call `SDL_SetMainReady()` under `#ifdef __ANDROID__` before `SDL_Init`, so a pure HL/native binary (not wrapped in an Android Activity) doesn't trip SDL's `SDL_main` check.

Both are guarded by GL ES / `__ANDROID__` and have no effect on desktop targets. Rebased onto the current SDL3 master.

### DIFF
--- a/libs/sdl/gl.c
+++ b/libs/sdl/gl.c
@@ -46,6 +46,7 @@
 #	define glPolygonMode(face,mode) if( mode != 0x1B02 ) ES_NOT_SUPPORTED
 #	define glGetQueryObjectiv glGetQueryObjectuiv
 #	define glClearDepth glClearDepthf
+#	define glColorMaski(...) ES_NOT_SUPPORTED
 #endif
 
 #if !defined(HL_CONSOLE) && !defined(GL_IMPORT)
--- a/libs/sdl/sdl.c
+++ b/libs/sdl/sdl.c
@@ -7,6 +7,12 @@
 #include <SDL3/SDL.h>
 #include <SDL3/SDL_gamepad.h>
 
+#ifdef HL_ANDROID
+// SDL3 no longer pulls SDL_main.h (where SDL_SetMainReady lives) in via SDL.h.
+#	define SDL_MAIN_HANDLED
+#	include <SDL3/SDL_main.h>
+#endif
+
 #if defined (HL_IOS) || defined(HL_TVOS)
 #	include <OpenGLES/ES3/gl.h>
 #	include <OpenGLES/ES3/glext.h>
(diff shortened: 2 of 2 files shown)


Pign:
Fixed the `build-android` failure: under SDL3 `SDL_SetMainReady()` is declared in `SDL_main.h`, which `<SDL3/SDL.h>` doesn't pull in. Added `#include <SDL3/SDL_main.h>` (with `SDL_MAIN_HANDLED`, so it doesn't redefine `main` in this shared library) guarded by `__ANDROID__`. Force-pushed the fix.

yuxiaomao:
I think it would be better to use `HL_ANDROID` instead of `__ANDROID__`, and reduce comment when possible

tobil4sk:
> sdl.c: call SDL_SetMainReady() under #ifdef __ANDROID__ before SDL_Init, so a pure HL/native binary (not wrapped in an Android Activity) doesn't trip SDL's SDL_main check.

Could you add a link to a section in the sdl3 migration guide that verifies that this is the correct fix?

Pign:
Done, switched to `HL_ANDROID` and trimmed the comments.

For the correctness of the fix, two references:
- SDL3 migration guide, *SDL_main.h* section: "SDL_main.h is now a header-only library **and not included by SDL.h anymore**." — this is why the explicit `#include <SDL3/SDL_main.h>` is now required. https://github.com/libsdl-org/SDL/blob/main/docs/README-migration.md#sdl_mainh
- `SDL_SetMainReady` docs: "Circumvent failure of `SDL_Init()` when not using `SDL_main()` as an entry point… it is necessary to define `SDL_MAIN_HANDLED` before including SDL.h." — exactly the pure-HL/native case here. https://wiki.libsdl.org/SDL3/SDL_SetMainReady

Pign:
Note: the Windows build failures on this run are infrastructure, not this change — the "Install: Required Dev Packages" step failed to download SDL (`curl (28) Failed to connect to www.libsdl.org port 443`) before any compilation. `build-android` and all Linux/macOS builds pass. I don't have rights to re-run the failed jobs; they should pass on a re-run.

yuxiaomao:
I don't use Android myself but it seems fine, so I'll merge

PR #953, opened 2026-07-01
ui_win: fix invocation of sentinel callback

Apprentice-Alchemist:
Fixes the issue described by @yuxiaomao in https://github.com/HaxeFoundation/hashlink/pull/902#issuecomment-4853286625

To simulate a call the stack (prior to pushing the return address) needs to be aligned to 16 bytes to respect the x64 calling convention, and only the return address should be pushed.

### DIFF
--- a/libs/ui/ui_win.c
+++ b/libs/ui/ui_win.c
@@ -228,14 +228,14 @@ static void sentinel_loop( vsentinel *s ) {
 				// simulate a call
 #				ifdef HL_64
 				int_val* rsp = (int_val*)regs.Rsp;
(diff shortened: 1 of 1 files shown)


yuxiaomao:
Thank you! It seems also useful even if we don't use unwind, to get a proper behavior in my sample (I don't know why timeout still works sometimes in real game project tho).

I also discovered that, timeout can be re-triggered if I drag and move UI pop-up for a while (I have heaps timeout set to 1 which normally should not be the case). On the second timeout it does not show any stack when using unwind (show full stack without unwind), but in both case the game / UI windows hang in a strange way.

PR #960, opened 2026-07-19
SDL3 File Dialogs

nspitko:
This PR adds support for File/Folder open/save dialogs via SDL3. These are platform agnostic async calls that should function consistently across all supported targets, and use modern APIs that should match the user's OS.

The implementation here is a bit icky in a few spots as we need to jump through some hoops to get strings in and out of HL. I chose to hide as much of this complexity in sdl.hx as possible, but arguments can be made for shifting it up or down the call tree. I don't think it matters much where this code lives as long as it's not user facing, so the important bit is the sdl.hx interface which exposes `FileDialogCallback` and `DialogFileFilter`.

Some notes:
- SDL will always trigger the callback, even if the user cancels or the function fails due to other reasons (eg, bad filter data, etc). With this contract it's safe to leave cleanup to `FileDialogCallback` in sdl.c and not worry about leaking memory or gc roots.
- Zeta assisted on the gc root code, but it's worth getting a second set of eyes on the implementation since I blindly ran with his suggestion; I don't fully understand the entire memory ownership model of the gc and it may not make sense the way I implemented it. It appears to work in my testing.
- I really wish we could pass strings through HL without the intermediate hl.bytes stage, but I couldn't find a way to do this.

### Use cases
- Currently, my in-house tools for heaps heavily lean on an older version of this code for save/load/selection on linux.
- Hide likely also needs this if we ever want it to work reasonably well outside of windows.

### Why not use UI.hdll?
Currently there are no other options for file dialogs in HL outside of windows. `ui.hdll` has some vague hints towards being portable but there's no work currently done on other platforms, and maintaining versions across all platforms is a non-trivial endeavor, especially on linux where there are a large number of windowing systems with subtly different APIs. SDL3 has already done this work for us, and is likely in use for window creation on non-windows platforms anyway.

### DIFF
--- a/libs/sdl/sdl.c
+++ b/libs/sdl/sdl.c
@@ -1203,7 +1203,7 @@ void FileDialogCallback(void *userdata, const char* const *filelist, int filter)
 	dialog_data *data = (dialog_data*)userdata;
 	int count = 0;
 
-	varray *array;
+	varray *array = NULL;
 	if( filelist ) {
 		const char * const *p = filelist;
 		while(*p++)
@@ -1218,8 +1218,6 @@ void FileDialogCallback(void *userdata, const char* const *filelist, int filter)
 			memcpy(bytes, filelist[i], len);
 			array_ptr[i] = bytes;
 		}
-	} else {
-		array = hl_alloc_array(&hlt_bytes, 0 );
 	}
 
 	hl_call1( void, data->closure, varray*, array );
--- a/libs/sdl/sdl.c
+++ b/libs/sdl/sdl.c
@@ -1163,14 +1163,12 @@ typedef struct {
 	SDL_DialogFileFilter* filters;
 } dialog_data;
 
-dialog_data* CreateFileDialogData( vclosure *callback, varray *filters )
-{
+dialog_data* CreateFileDialogData( vclosure *callback, varray *filters ) {
 	dialog_data *data = malloc( sizeof( dialog_data ) );
 	data->closure_store = malloc(sizeof(vclosure*));
 	*data->closure_store = callback;
 
-	if( filters && filters->size > 0 )
-	{
+	if( filters && filters->size > 0 ) {
 		SDL_DialogFileFilter *sdl_filters = (SDL_DialogFileFilter *)malloc(sizeof( SDL_DialogFileFilter ) * filters->size );
 
 		for(int i=0;i<filters->size;i++) {
--- a/libs/sdl/sdl.c
+++ b/libs/sdl/sdl.c
@@ -1160,20 +1160,19 @@ DEFINE_PRIM(_BYTES, get_error, _NO_ARG);
 // SDL Dialogs API
 void FileDialogCallback(void *userdata, const char* const *filelist, int filter) {
 	// these callbacks may come via threads on some platforms
-	bool on_thread = !hl_get_thread();
-	if( on_thread )
-	{
-		vdynamic* ctx;
+	bool on_unregistered_thread = !hl_get_thread();
+	if( on_unregistered_thread ) {
+		vdynamic *ctx;
 		hl_register_thread(&ctx);
 	}
 
-    vclosure* vcallback = (vclosure*)userdata;
+	vclosure **closure_store = userdata;
+	vclosure *vcallback = *closure_store;
 
 	int count = 0;
 
 	varray *array;
-	if( filelist )
-	{
+	if( filelist ) {
 		const char * const *p = filelist;
 		while(*p++)
 			count++;
@@ -1187,16 +1186,15 @@ void FileDialogCallback(void *userdata, const char* const *filelist, int filter)
 			memcpy(bytes, filelist[i], len);
 			data[i] = bytes;
 		}
-	}
-	else 
-	{
+	} else {
 		array = hl_alloc_array(&hlt_bytes, 0 );
 	}
 
 	hl_call1( void, vcallback, varray*, array );
-	hl_remove_root( &vcallback );
+	hl_remove_root( closure_store );
+	free( closure_store );
 
-	if( on_thread )
+	if( on_unregistered_thread )
 		hl_unregister_thread();
 }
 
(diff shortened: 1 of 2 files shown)


libs/sdl/sdl.c
```diff
@@ -1156,3 +1156,102 @@ DEFINE_PRIM(_ARR, get_display_modes, _I32);
...
+	vclosure **closure_store = malloc(sizeof(vclosure*));
+	*closure_store = callback;
+	const char *location = default_location ? hl_to_utf8(default_location->bytes) : NULL;
+	
+	hl_add_root(closure_store);
+	
+	SDL_DialogFileFilter *sdlfilters = CreateFileDialogFilters( filters );
+	SDL_ShowOpenFileDialog( FileDialogCallback, closure_store, window, sdlfilters, filters ? filters->size : 0, location, allow_many );
+
+	free( sdlfilters );
```
yuxiaomao: Seems that we should not free filters until callback?

https://wiki.libsdl.org/SDL3/SDL_ShowOpenFileDialog
filters: a list of filters, may be NULL. Not all platforms support this option, and platforms that do support it may allow the user to ignore the filters. If non-NULL, it must remain valid at least until the callback is invoked.

yuxiaomao: If we need to keep the filters until callback, we might need to do `add_root` on a struct / context that contains both callback and filters?

libs/sdl/sdl.c
```diff
@@ -1156,3 +1156,102 @@ DEFINE_PRIM(_ARR, get_display_modes, _I32);
...
 DEFINE_PRIM(_BYTES, get_error, _NO_ARG);
+
+// SDL Dialogs API
+void FileDialogCallback(void *userdata, const char* const *filelist, int filter) {
+	// these callbacks may come via threads on some platforms
+	bool on_unregistered_thread = !hl_get_thread();
+	if( on_unregistered_thread ) {
+		vdynamic *ctx;
+		hl_register_thread(&ctx);
+	}
```
yuxiaomao, line 1201: I'm not sure it's safe to call a Haxe function from a non-Haxe thread, especially if the callback involves complex logic or throws an exception. Would it be possible to dispatch the result back to the Haxe thread?

libs/sdl/sdl/Sdl.hx
```diff
@@ -277,6 +283,84 @@ class Sdl {
 	private static function _getJoysticks() : hl.NativeArray<Int> {
 		return null;
 	}
+
+	//
+	// SDL3 Dialogs API
+	//
+
+	private static function processFilters(filters: Array<DialogFileFilter> = null)
+	{
```
yuxiaomao: I'll certainly complains about the formatting in the files x) But I can probably fix that by my own if it's the only remaining things

libs/sdl/sdl.c
nspitko: From a calling convention standpoint I don't have a preference here, but I have no idea how to do that. Can you point me towards an example where we do this elsewhere in HL?

For what it's worth, I've done this in a few other hdlls and never had any issues with it, but that doesn't mean I haven't just been dodging landmines.

yuxiaomao: Double thinking about this, I think this solution is probably the simplest, especially if that worked for you.

I had in mind what I've done for XBS, where most of the API calls are async. The callback results are stored in a queue (in C, managed by the SDK in this case) and haxe side had an MainLoop.Event to check if any callback need to be called. This seems complex for only open/save file here.

libs/sdl/sdl/Sdl.hx
nspitko: Old habits die hard... I thought I cleaned them all up but guess not. Should be fixed now (unless there are more)

libs/sdl/sdl.c
nspitko: I added a struct. I THINK we only need to add_root on the closure_store here, the GC shouldn't know/care about mallocs in hlsdl, but happy to be corrected if there's a need to root the whole struct.

libs/sdl/sdl.c
```diff
@@ -1156,3 +1156,106 @@ DEFINE_PRIM(_ARR, get_display_modes, _I32);
...
+		vbyte **array_ptr = (vbyte **)hl_aptr(array, vbyte*);
+
+		for(int i = 0; i < count; i++) {
+			size_t len = strlen(filelist[i]) + 1; // Include the null terminator
+			vbyte *bytes = hl_alloc_bytes( len );
+			memcpy(bytes, filelist[i], len);
+			array_ptr[i] = bytes;
+		}
+	} else {
+		array = hl_alloc_array(&hlt_bytes, 0 );
```
yuxiaomao: When I'm trying to call with a filter on Windows, it return immediatly an empty array without open any window (works normally if no filters specifield). From C side I saw that `filelist` is NULL so error occurs, checking the error string it says 

> Invalid dialog file filters: Invalid character in pattern (Only [a-zA-Z0-9_.-] allowed, or a single *

I think it would be good if we can at least give some information to the haxe side about the error, instead of returning empty list here (return NULL or throw?).

```haxe
sdl.Sdl.showOpenFileDialog(function(arr) {
		trace("Open " + arr);
	}, null, [{ name : "hxml", pattern : "*.hxml" }]
);
```

libs/sdl/sdl.c
yuxiaomao: GC should not know about mallocs, but it seems that Filters' name/pattern still reference to hl.Bytes. We might need to avoid GC on them (either copy bytes with malloc; or make the struct gc managed pointer + add raw varray into it)

nspitko: Did a quick cleanup of this and the memory ownership a bit, strdup should work fine here but I can malloc instead if you prefer.

There's still an edge case where passing null as a filter var crashes hlsdl, need to handle that somewhere but I'll probably just throw in haxe before sending bad data unless we have a general policy that hl libs should be hardened at the C side.

yuxiaomao: strdup is good.
I'm not sure what do you reference by filter var.
hl libs can assume that data is validated by haxe, and haxe API can also assume that the call argument nullable or not is respected (if the function signature is explicit). So just let null filter name/pattern throw in haxe when performing `.toUtf8` is acceptable from my point of view.

libs/sdl/sdl.c
nspitko: This one's kinda tricky since we can't validate filters before the callback (the SDL internal `validate_filters` function is not exposed to us), so showOpenFileDialog can only ever be void. 

What I've done is to pass the NULL state into haxe, and `throw getError()` in that case. It works here, but I'm not sure there's any way to catch these since they fire in a thread, so I'm not super happy with the solution. The other option is we just pass null through and leave it to the user to read the API, which is more "correct" but probably not intuitive.

yuxiaomao: I don't think it's only about filters, errors can come from elsewhere, so it's good that at least we return NULL from C side.
Yes I agree with you it's not trivial. Another related problem is if we throw inside the callback (either inside SDL.hx or user code), `hl_remove_root` and `free` are never called.

libs/sdl/sdl.c
yuxiaomao: I told some dev this workaround, he tried (on some GPU code) but also found that when it's actually from another thread, if the callback throw, it will access violation in `hl_throw`.
On windows OpenFileDialog is called from the same thread so I didn't check yet

yuxiaomao: The solution found is to try-catch around user callback https://github.com/HeapsIO/heaps/blob/14264fdb914b59620b0ad24714eb76fced410812/h3d/impl/DX12Driver.hx#L3617-L3628

PR #964, opened 2026-07-27
Win: Add no console by default and attach later

yuxiaomao:
It seems that sometimes for a game release, we don't want the cmd window pop-up. This PR:
- Added `wWinMain` in main.c so one can compile with `subsystem:windows`
- Added `ui_create_console`, `ui_attach_console` to enable the console when we need it (API in haxe)
- Activated for the configuration `ReleaseDX12Agility`

The downside (?) is that the console window will look like a normal window instead of a cmd window.

Haxe side API:

```haxe
	#if (hl_ver >= version("1.16.0"))
	@:hlNative("?ui", "ui_create_console")
	public static function createConsole():Bool {
		return false;
	}

	@:hlNative("?ui", "ui_attach_console")
	public static function attachConsole(pid:Int):Bool {
		return false;
	}
	#end

```

For a better experience with debugger I think the expected usage is :

```haxe
if( !hl.Api.hasDebugger() ) {
	hl.UI.createConsole();
}
```

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -325,7 +325,7 @@ int main(int argc, pchar *argv[]) {
 	return 0;
 }
 
-#ifdef HL_WIN
+#if (defined(HL_WIN_DESKTOP) && !defined(_CONSOLE)) || defined(HL_XBS)
 int WINAPI wWinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, PWSTR lpCmdLine, int nCmdShow) {
 	return wmain(__argc, __wargv);
 }
--- a/src/main.c
+++ b/src/main.c
@@ -325,3 +325,8 @@ int main(int argc, pchar *argv[]) {
 	return 0;
 }
 
+#ifdef HL_WIN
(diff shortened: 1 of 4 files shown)


yuxiaomao:
Another pb when the debugger is attached (and as we don't do createConsole to keep the trace inside the debugger) is that now every `Sys.command` will spawn its own console, which is not ideal

src/main.c
```diff
@@ -325,3 +325,8 @@ int main(int argc, pchar *argv[]) {
 	return 0;
 }
 
+#ifdef HL_WIN
```
tobil4sk: Maybe it would be good to have `&& !defined(_CONSOLE)`, so people can still compile it for the console subsystem. This would match `hlc_main.c`:

https://github.com/HaxeFoundation/hashlink/blob/1fa6e80b95f2b9fd6668e9577bcb97369dcce78b/src/hlc_main.c#L178

yuxiaomao: Yes we can keep the same (but it seems still compile when choosing console subsystem).
HL_XBS can use the same too because I copied the function from HL_XBS, so I put the same condition here.

PR #967, opened 2026-07-31
Update vendored sqlite from 3.15.1 to 3.53.4

KamaleiZestri:
Took me a while to figure out why UPSERT commands worked on linux but not windows.


yuxiaomao:
The previous lib was 10 years ago, I suppose indeed it's the time for update x)

PR #971, opened 2026-08-06
[dx] adaptive dark mode

ournuu:
I think it would be better to match the SDL backend's behavior instead of enabling dark mode manually


ournuu:
Please notice me!!

PR #972, opened 2026-08-10
Command Queue removed from static_driver

julesvannier:


libs/directx/dx/Dx12.hx
```diff
@@ -1751,26 +1747,20 @@ class Dx12 {
...
 	}
 
-	public static function signal( fence : Fence, value : Int64 ) {
-	}
-
-	public static function wait( fence : Fence, value : Int64 ) {
-	}
-
-	public static function present( vsync : Bool ) {
+	public static function present( directQueue : CommandQueue, vsync : Bool ) {
```
TothBenoit: To correctly mimic the DirectX 12 API, I would move present, suspend, resume and getTimestampFrequency into the CommandQueue class.

TothBenoit (changes requested):

libs/directx/dx/Dx12.hx
julesvannier: done

PR #978, opened 2026-09-04
Link Apple's OpenGL framework on macOS

darmie:
`find_package(OpenGL)` reports a Homebrew Mesa `libGL` ahead of the framework when `CMAKE_FIND_FRAMEWORK` is `LAST`. `sdl.hdll` then links and loads fine, and every `gl*` call goes to a Mesa with no current context.

Symptoms: `glGetString` returns NULL, `glCreateShader` returns 0. `Window.testGL()` checks neither, so the first visible failure is a segfault in `glShaderSource` on a null shader.

Only Apple's framework can serve the CGL context SDL creates, so ask for it by name.

Verified on macOS arm64 (SDL 3.4.16), configuring with `-DCMAKE_FIND_FRAMEWORK=LAST` and Mesa installed:

- before: `sdl.hdll` links `/opt/homebrew/opt/mesa/lib/libGL.1.dylib`; `GL_VERSION` is null, `createShader` null
- after: links `OpenGL.framework`; `GL_VERSION` = `4.1 Metal - 90.5`, `GLSL` = `4.10`, `createShader` = 1, Heaps/HaxeUI app renders

### DIFF
--- a/libs/sdl/CMakeLists.txt
+++ b/libs/sdl/CMakeLists.txt
@@ -62,9 +62,19 @@ if((APPLE OR UNIX) AND NOT ANDROID)
         PRIVATE
         ${OPENGL_INCLUDE_DIR}
     )
+    if(APPLE)
+        # Only Apple's framework can serve the CGL context SDL creates.
+        # find_package(OpenGL) picks a Homebrew Mesa libGL ahead of it when
+        # CMAKE_FIND_FRAMEWORK is LAST; that links fine and then every gl*
+        # call returns 0/NULL.
+        find_library(HL_APPLE_OPENGL_FRAMEWORK OpenGL REQUIRED)
+        set(HL_SDL_GL_LIBRARY ${HL_APPLE_OPENGL_FRAMEWORK})
+    else()
+        set(HL_SDL_GL_LIBRARY ${OPENGL_gl_LIBRARY})
+    endif()
     target_link_libraries(sdl.hdll
         libhl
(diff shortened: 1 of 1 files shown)


tobil4sk:
If a user has passed in `CMAKE_FIND_FRAMEWORK=LAST` then I would expect the homebrew library to take priority over the framework. Why would we want to override that?

darmie:
@tobil4sk  then perhaps we need a way to override Hashlink's Cmake config globally without altering existing global CMAKE environment flags.

tobil4sk:
Where exactly does `-DCMAKE_FIND_FRAMEWORK=LAST` get set? I don't think it is coming from hashlink's cmakelists.txt.

I think you can control where find package will look using variables like `OpenGL_DIR`. See the docs: https://cmake.org/cmake/help/latest/command/find_package.html.

darmie:
> Where exactly does `-DCMAKE_FIND_FRAMEWORK=LAST` get set? I don't think it is coming from hashlink's cmakelists.txt.
> 
> I think you can control where find package will look using variables like `OpenGL_DIR`. See the docs: https://cmake.org/cmake/help/latest/command/find_package.html.

Thanks, I'll verify this.

darmie:
@tobil4sk Thanks for the pointer, the issue is confirmed to be environment specific.  And it has been resolved by honoring Hashlink's cmake defaults and removing the problematic mesa opengl installed via homebrew. 

Closing this PR.

PR #979, opened 2026-09-08
[ui] Added isFolder to ui_choose_file options

EspeuteClement:
This add the ability to specify that we want a folder instead of a file when prompting the user with the native choose file dialog.

This add a new utils.cpp file, because having a prompt to choose a folder using windows filebrowser need to use a COM based interface, and these are very tricky to use in pure C. Using C++ allows us to use the api as intended, and we make a simple wrapper around that that can be then called from the C side.

### DIFF
--- a/libs/ui/utils.cpp
+++ b/libs/ui/utils.cpp
@@ -1,0 +1,55 @@
+/* Functions that need cpp support for ui_win.c */
+
+#include <shobjidl.h>
+#include <windows.h>
+
+/** 
+ * Returns false if an error happened while choosing the folder, else returns true and outBuffer
+ * contains the file path
+ * **/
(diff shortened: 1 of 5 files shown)


ncannasse:
ok pour moi, mais build cmake à revoir

EspeuteClement:
Cmake build should now be correct

libs/ui/utils.cpp
```diff
@@ -0,0 +1,55 @@
+/* Functions that need cpp support for ui_win.c */
+
+#include <shobjidl.h>
+#include <Windows.h>
```
tobil4sk: Lower case `windows.h` would be more compatible with mingw (cross compiling from linux)

EspeuteClement: Fixed in the latest push, thanks

